PR #22, opened 2023-06-02
Deduplicate bsp file, move fusion file, path handling

mrosseel:


brickbots (approved):
This all looks great, thank you for pushing on this much-needed cleanup work!

PR #23, opened 2023-06-04
#2 Keyboard server

mrosseel:
* added a keyboard server, which serves a basic webpage representing the pifinders' button interface. Supports 'alt' and 'long' modifiers. This adds the simple 'bottle' dependency for serving the website, looked into socket but didn't seem the best fit, too much code.
* small fix to keyboard_local so that it warns the user that it doesn't work on the pi. I haven't removed it because it's great that this just works on mac (and probably win/linux but haven't tested)
* reduced PIL logging in debug mode, because it swamps the logs
* extracted camera and keyboard from the fake hardware switch, because ultimately for testing you can have a working keyboard but no camera, everything but the imu, etc.

See [example](https://discord.com/channels/1087556380724052059/1092219845954781274/1114651183018152116)

Long term I might try to include the screen into the webpage so it can be run completely headless

### DIFF
--- a/python/PiFinder/keyboard_local.py
+++ b/python/PiFinder/keyboard_local.py
@@ -1,11 +1,14 @@
 import time
 from PiFinder.keyboard_interface import KeyboardInterface
 import logging
-from PyHotKey import Key, keyboard_manager as manager
-
 
 class KeyboardLocal(KeyboardInterface):
     def __init__(self, q):
+        try:
+            from PyHotKey import Key, keyboard_manager as manager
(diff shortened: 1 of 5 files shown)


brickbots:
This is also nice work!  Bottle is great for this and much better than writing all the logic for a reliable web server.  I'm going to merge this, and the image serving portion.

Breaking out the various options (keyboard / display / etc) is a good idea here.  There are probably a lot of interesting combinations that could be useful for folks to build on.

PR #24, opened 2023-06-06
#1 make POSS/SDSS Fov prettier

mrosseel:
This is my proposal to make the image viewer a bit nicer by removing the black rectangles which provide contrast for the POSS / 1.0 text showing image catalog and fov. They are now replaced with outlined text. The fov has an added degree symbol and is full red to make it clearer to see (more important than the image catalog). In the image below, POSS has a shadow, the fov has an outline. Please adjust to taste.

Screenshot 2023-06-06 at 12 49 53

Would it be an idea to convert < 1 degree into minutes using the " symbol? Not sure what's most helpful at the eyepiece. Maybe ultimately the eyepiece fl.

NOTE: should probably be merged before the other PR

### DIFF
--- a/python/PiFinder/cat_images.py
+++ b/python/PiFinder/cat_images.py
@@ -4,8 +4,6 @@
 This module is used at runtime
 to handle catalog image loading
 """
-from queue import Queue
-import requests
 import sqlite3
 import os
 from PIL import Image, ImageChops, ImageDraw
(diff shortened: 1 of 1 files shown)


brickbots:
The outlined text looks great!  Good question about fractional degrees vs. arc minutes vs. eyepiece focal length.  We can decide when the eyepiece stuff gets implemented 👍

PR #26, opened 2023-06-08
#1 added SAC Asterisms + catalog refactoring

mrosseel:
Example of title scrolling:
https://github.com/brickbots/PiFinder/assets/1465310/944ce5bf-8664-45da-b01c-0aee88eccab0

Need some feedback for speed, stopping duration at start, does this need to be repeated forever, etc.


brickbots:
This looks great!  Merging!  I'll use it a bit and then comment on the scrolling speed, but my initial thoughts are we might want to make this 'wrap' so that it's always moving at a constant speed?

I'm excited to look through the Asterisms catalog 👍

mrosseel:
Yes, scrolling needs some user feedback. In case you missed it, there's a
config entry to adjust scrolling speed so you can test/adjust this.

PR #31, opened 2023-06-20
make the remote control website a bit easier on the eye

mrosseel:
* added title
* made keypad sit right of the screen, as in real life + easier to see everything
* change colors to make it darker

Screenshot 2023-06-20 at 15 00 21

### DIFF
--- a/python/index.tpl
+++ b/python/index.tpl
@@ -1,37 +1,52 @@
 <!DOCTYPE html>
 <html>
+<head>
+<style>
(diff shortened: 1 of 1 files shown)


brickbots:
This keeps getting better!  At least one person asked about a headless version that used a phone/tablet.  I'm not sure how many people might use it like this, but I'm tempted to build up a unit without keypad and see how usable it is!

PR #41, opened 2023-06-23
Stop flickering of html page when server is down

mrosseel:


brickbots (commented):
My Javascript is not fantastic, but this looks like a good solution implemented in the JS form!

PR #42, opened 2023-06-23
Pushing the B button crashes when a log is present

mrosseel:
There was an incorrect variable name in an unfrequently used code path which crashed during field usage.


brickbots:
Glad you were able to track this one down.  Since it requires having logged the object previously, it could have been really hard to troubleshoot.  Thank you!!

PR #47, opened 2023-06-27
(1) Catalog refactor

mrosseel:
Refactor the catalog.py class

- cleaner separation between logic and UI
- objects are now in dicts so no longer needs to search for an object (was slow for the higher numbers)
- remembers which object was selected in each catalog, so you can more freely switch
- long C now goes backwards in the catalog list in case you overshot your target catalog

TODO:

- [x] going from locate to catalog should select the object selected in locate
- [x] up/down bug where it goes from M2 to --- after trip to locate
- [x] cleanup debug logging
- [x] split up the ui_utils.py class?


mrosseel:
This can be merged, up to you if it's stable enough to go into the next release. I've tested most of what I know but maybe there's some stuff untested - e.g. observing lists.

python/PiFinder/imu_pi.py
```diff
@@ -20,8 +20,8 @@ class Imu:
     def __init__(self):
         i2c = board.I2C()
         self.sensor = adafruit_bno055.BNO055_I2C(i2c)
-        # self.sensor.mode = adafruit_bno055.IMUPLUS_MODE
-        self.sensor.mode = adafruit_bno055.NDOF_MODE
+        self.sensor.mode = adafruit_bno055.IMUPLUS_MODE
```
brickbots, line 23: I'm still evaluating both of these options... did you see  any difference between them in your usage?

requirements.txt
```diff
@@ -1,7 +1,7 @@
 adafruit-blinka==8.12.0
 adafruit-circuitpython-bno055
 luma.oled==3.12.0
-pillow==10.0.0
+pillow
```
brickbots: I've just tested version 9.5.0 on the hardware and emulated version.  I'll set this in the requirements file

docs/dev_guide.md
```diff
@@ -1,3 +1,47 @@
 # PiFinder software development guide
```
brickbots, line 1: Thank you SOOO MUCH for getting this starts 🥇

PR #52, opened 2023-07-05
Update BOM.md

JanKok:
I didn't fimd a "Diptonics DTS63K 1nm" but I did find a Diptronics part and an Apem part which I assume were what you intended. Supplied some URLs and corrected some typos,

BTW thanks for publishing this very interesting project! - Jan


brickbots:
Thanks @JanKok for the updates/cleanup of the BOM!!

PR #53, opened 2023-07-06
cleanup/fixes of double star textlayout changes

mrosseel:


brickbots:
Thanks @mrosseel !  This looks good, I'm a bit behind in branching the release candidate, so this will make it into that branch for testing :+1:

PR #54, opened 2023-07-08
Global search, second try

mrosseel:
- push objects near your current position
- added random keystroke generator to check for UI crashes (found 2 and solved)
- renamed CatalogTracker.current to CatalogTracker.current_catalog

TODO

- [x] make sure searching works


python/PiFinder/main.py
```diff
@@ -709,7 +710,7 @@ def main(script_name=None):
         from PiFinder import keyboard_server as keyboard
 
     if args.log:
-        datenow = datetime.now()
+        datenow = datetime.datetime.now()
         filehandler = f"PiFinder-{datenow:%Y%m%d-%H_%M_%S}.log"
```
brickbots, line 714: I know this line has been here for a bit, but in the field the PiFinder likely won't have the correct date time.  In other areas where a distinct filename is needed, the UID of the session is used in place of the date time stamp.

It makes the files a bit harder to ultimately use, as they are just random strings, but at least all the files generated from the same execution of the software have the same filename prefix.  

After merging, I'll adjust this line to use the session UID unless you have some other idea on how to deal with the lack of solid date and timekeeping in the field without internet?

python/PiFinder/ui/catalog.py
```diff
@@ -73,6 +73,12 @@ class UICatalog(UIModule):
             "options": ["CANCEL", 5, 10, 15, 20],
             "callback": "push_near",
         },
+        "Push All Near": {
```
brickbots, line 76: I think these two option descriptions need some work..... but I don't know what to do to make them really intuitive with the limited text length.  Ideally they would be something like:

* Find nearby objects in current catalog
* Find nearby objects in all catalogs

I'll think about this a bit, but if you have any suggestions let me know :-)

brickbots (approved):
This all looks good and in my quick testing it seems to work.  I'm going to merge this through to the release candidate for under-the-stars testing!

python/PiFinder/main.py
mrosseel: isn't the time coming from the GPS?

python/PiFinder/ui/catalog.py
mrosseel: agreed, not sure what's better.
Near in Cat
Near in All
?

python/PiFinder/main.py
brickbots: It could... although I've not setup GPSd to set the system time.  Even if it was setting the system time, it's very unlikely there would be a fix by the time this code runs during the first minute or so of system startup.  It is kind of a bummer not to have reliable date/time from system boot... I've definitely been spoiled by the era of network connected everything!

python/PiFinder/ui/catalog.py
brickbots: Thank you!!  That's better than the current state, let's go with that for now!

PR #58, opened 2023-07-16
1 - Add pagination to scrolling

mrosseel:
- use RGBA whenever possible (autocontrast doesn't support it)
- catch errors in threads for cleaner debugging during dev
- web page now checks if /image is back, avoids manual refreshing.
- add a scroll bar indicating your current position in the description


mrosseel:
Screenshot 2023-07-18 at 23 49 19

mrosseel:
@brickbots if you're not comfortable with the RGBA changes we can discuss.

PR #63, opened 2023-07-24
New database structure

mrosseel:
TODO: convert all catalogs into the new DB format

- [x] NGC
- [x] M
- [x] Caldwell
- [x] Collinder
- [x] TAAS200
- [x] SaA
- [x] SaM


python/PiFinder/catalogs.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import sqlite3
 import time
+from db.db import Database, ObjectsDatabase
```
brickbots, line 4: Is this a new file/module yet to be checked in?

brickbots (commented):
Overall looking great!  I didn't see the db module file, but perhaps I've missed it from another PR?

python/PiFinder/catalogs.py
mrosseel: there were indeed some essential pieces missing :)

brickbots (commented):
I really appreciate your efforts to move as much to objects as possible.  It's going to help future maintenance enormously!

steamraven:
I see the CompositeObject just forwards attributes to the underlying dict.  I am worried that if we want to use Type Hints in the future, this will be hard to check against.  Can we use an explicit Dataclass for something like this?

steamraven:
Or, what about using TypedDicts? Since most of the other code in PiFinder uses dicts, TypedDicts would match the style of the rest of the code, allow interop, and still get the benefit of type checking.

mrosseel:
> Or, what about using TypedDicts? Since most of the other code in PiFinder uses dicts, TypedDicts would match the style of the rest of the code, allow interop, and still get the benefit of type checking.

could be an improvement, but the dicts are coming directly from the DB which is kinda typed so I don't see much value atm. Would prefer to finish this PR first, it's already quite big.

PR #76, opened 2023-08-15
Split server from keyboard

mrosseel:
The webpage running on 8080 should be more generic than just a keyboard, so it is now a server component no longer pretending to be a keyboard.
Added GSP lock button to push a GPS fix on the queue. GPS_QUEUE is not part of the keyboardinterface so would start getting awkward to add more things to the server.

### DIFF
--- a/python/PiFinder/main.py
+++ b/python/PiFinder/main.py
@@ -19,7 +19,7 @@ import sys
 import logging
 import argparse
 from pathlib import Path
-from PIL import Image, ImageDraw, ImageFont, ImageChops, ImageOps
+from PIL import Image, ImageOps
 from multiprocessing import Process, Queue
 from multiprocessing.managers import BaseManager
 from timezonefinder import TimezoneFinder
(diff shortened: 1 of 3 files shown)


brickbots (approved):
Thanks for getting this ready for some new development!  Here are some potential suggestions for this PR, but they can come in a subsequent one as well:

* Move the server to port 80.  I think this web interface is going to become a valuable configuration tool, so let's make it a first class citizen
* Make it run all the time, rather than require a parameter on startup... for the same reason of having user friendly way to set certain config items
* Expand the GPS end point to accept lat/long/alt and time/date as part of the payload.  This *could* allow the use of a PiFinder without GPS if someone was willing to use the web interface to initialize it with the location/time&date.  We might never really want to support this, but it's an interesting idea.

Great stuff here, so ping me if you want to merge as is!

mrosseel:
all valid points, would prefer to get this merged as-is though, not sure of the scope of all the changes yet. Might need for instance a tabbed interface with one tab for config, another for the UI emulation, another for debug information, etc.

PR #80, opened 2023-08-26
Update object image code to use new DB structure

brickbots:
Also include fix to logging system crash I found while testing.


python/PiFinder/setup.py
```diff
@@ -163,8 +164,67 @@ def insert_catalog_max_sequence(catalog_name):
     conn.commit()
 
 
+def resolve_object_images():
+    # This is the list of catalogs to search for
+    # objects to match against image names
+    resolution_priority = [
```
mrosseel, line 170: nothing wrong with this for now, but would like to have an extra column in the catalog table with the priority for that catalog

python/PiFinder/setup.py
```diff
@@ -163,8 +164,67 @@ def insert_catalog_max_sequence(catalog_name):
...
+            if catalog_check:
+                # Found a match!
+                resolved_name = f"{catalog_code}{catalog_check['sequence']}"
+                break
+
+        if resolved_name == None:
+            # Didn't find a name in the priority list
+            # Pick one from the remainders, but sort
+            # by catalog code so this is deterministic
+            catalog_entry = db_c.execute(
```
mrosseel, line 205: why not objects_db.get_catalog_object_by_object_id() ?

mrosseel (changes requested):

python/PiFinder/setup.py
```diff
@@ -163,8 +164,67 @@ def insert_catalog_max_sequence(catalog_name):
...
+    # objects to match against image names
+    resolution_priority = [
+        "NGC",
+        "IC",
+        "M",
+        "C",
+    ]
+    # load all objects in objects table
+    conn, db_c = objects_db.get_conn_cursor()
+    all_objects = db_c.execute(
```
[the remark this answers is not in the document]
brickbots: Indeed!  I've converted this over

python/PiFinder/setup.py
brickbots: I've changed this to use the insert order of the catalogs for priority.  I slightly adjusted the catalog import order and added a comment to remind future folks that the order is important

python/PiFinder/setup.py
brickbots: Rewrote this to just continue down the priority order until it finds one which currently yields the same output.

brickbots:
Thanks for the review @mrosseel, some great suggestions that caused me to rethink a bit and turn out better code 👍

PR #82, opened 2023-08-31
Bright star snap alignment

brickbots:
From preview screen:

- 'B' to start Bright Star Alignment
- Three brightest stars are highlighted
- Press the number key of the star to align to.  Or 0 to snap to center of frame.

Todo:

- [x] Change position of highlight/numbers for stars close to edge of image so they are always visible


brickbots:
@mrosseel When you get a chance, checkout this branch and let me know what you think!  Works fine with the normal debug image/method.

python/PiFinder/state.py
```diff
@@ -26,6 +27,20 @@ def __init__(self):
...
+        self.__solve_pixel = config.Config().get_option("solve_pixel")
+
+    def solve_pixel(self, screen_space=False):
+        """
+        solve_pixel is (Y,X) in camera image (512x512) space
+
+        if screen_space=True, this is returned as (X,Y) in screen (128x128) space
+        """
+        if screen_space:
+            return (int(self.__solve_pixel[1] / 4), int(self.__solve_pixel[0] / 4))
```
mrosseel, line 39: I've probably done this myself in some places, but at one point we'll have another screen resolution -
so maybe define this somewhere central, screen size, camera size, screen-camera ratio

python/PiFinder/ui/preview.py
```diff
@@ -113,31 +116,79 @@ def draw_reticle(self):
...
+                solve_pixel[1] + circ_rad,
             ]
             self.draw.arc(bbox, 20, 70, fill=self.colors.get(brightness))
             self.draw.arc(bbox, 110, 160, fill=self.colors.get(brightness))
             self.draw.arc(bbox, 200, 250, fill=self.colors.get(brightness))
             self.draw.arc(bbox, 290, 340, fill=self.colors.get(brightness))
 
+    def draw_star_selectors(self):
+        # Draw star selectors
+        if self.star_list.shape[0] > 3:
```
mrosseel, line 135: what if only one star can be seen, why wouldn't we want to center on it?

python/PiFinder/ui/preview.py
```diff
@@ -56,13 +60,6 @@ class UIPreview(UIModule):
             "options": ["Save", "Exit"],
             "callback": "save_exp",
         },
-        "Zoom View": {
```
mrosseel, line 59: Zoom View is handy when not in self-align mode, removing this implies we'll always be in self-align mode? If it works correctly why not of course. Or is the self-align mode to distinguish between just the preview, and the star-selecting preview?

python/PiFinder/ui/preview.py
```diff
@@ -169,7 +234,33 @@ def key_enter(self):
         self.command_queues["camera"].put("exp_save")
 
     def key_number(self, number):
-        if number == 0:
-            self.capture_count += 1
-            capture_imagepath = self.capture_prefix + f"_{self.capture_count :0>3}.png"
-            self.command_queues["camera"].put("save:" + capture_imagepath)
+        if self.align_mode:
+            if number == 0:
```
mrosseel, line 238: hard to discover, what does it even do, set the solve pixel to the center of the screen? Do we want this to be available? Might be confusing.

mrosseel (approved):
looks good, some small questions/nitpicks but merge away if you don't find anything valuable.

default_config.json
```diff
@@ -13,6 +13,7 @@
     "screen_direction": "right",
     "solver_debug": 0,
     "sleep_timeout": "30s",
+    "solve_pixel": [256, 256],
```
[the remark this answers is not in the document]
brickbots: That's pretty much it.  Either a restart with the finder still on the scope, or even perhaps a good shoe/dovetail fit that reproduces alignment between physical mounting events.

python/PiFinder/state.py
brickbots: Indeed!  Work for a future (maybe not too far) day....

python/PiFinder/ui/preview.py
brickbots: Good point, it's unlikely there would be one bright star without at least two others in the image, but no reason to limit it here.  I'll make that change.

python/PiFinder/ui/preview.py
brickbots: Once this code is merged the RA/DEC will always be reported based on the solve_pixel of the image, which by default will be the center pixel.  The 'B' key is just to enable the star selection feature in the preview, but once the star is selected that pixel becomes the new solve_pixel until it's reset.  The zero key while in star selection mode will reset to center.   Which reminds me I need to do some docs for this :-)

python/PiFinder/ui/preview.py
brickbots: Definitely need some documentation around this, but indeed it resets the solve position /reticle to the center.  The preview screen will show the reticle placed around the solve_pixel so it's a bit more intuitive than the code might indicate.

PR #83, opened 2023-09-03
Add a non-solving camera

mrosseel:


brickbots:
Will this module return a blank, or other non-star image to prevent any solving from happening?  Or is the intent to just never return an image, simulating a camera that never takes exposures?

mrosseel:
it should simulate an image not solving, not a faulty camera. It worked for
the testing I did with it but maybe it was doing the second thing

Mike

python/PiFinder/camera_none.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def get_cam_type(self) -> str:
+        return self.camType
+
+
+def get_images(shared_state, camera_image, command_queue, console_queue):
+    """
+    Instantiates the camera hardware
+    then calls the universal image loop
+    """
```
brickbots, line 58: I believe this current implementation will just cause the camera thread to exit.  The `get_images` function is called to start the image loop.  If you copy what is in the camera_debug.py for the `get_images` function, it has the minimum required to start and maintain the image loop.

Then you change the `capture` method of the `CameraNone` object to either return a blank image, a non-solving image from disk, or (my personal favorite) write a random dot generator return plausible, but not solvable, star images for the whole system to chew on.

brickbots (commented):
I like this for testing assumptions about having solvable images... Putting the lens cap on the camera might be a similar effect, but this one works on a PC too!

python/PiFinder/camera_none.py
mrosseel: will have another look at this later, but you're right it should do something more than exiting

brickbots (approved):

PR #84, opened 2023-09-05
Cleaned up and rationalized debug dumps

brickbots:
Ent-D is for debug dumps!  One key to rule them all......

* Add shared state, ui state, screenshot to debug dump
* Made sure raw and filtered camera output is saved
* Switch dump location to proper ~/PiFinder_data/solver_debug_dumps
* Fix screenshot code (Ent-0) as a side effect
* Remove camera image saving from other locations
* Remove exposure up/down/save from Preview (general cleanup)

Now, no matter what you are seeing/investigating Ent-D should save enough info really debug and to get the PiFinder back in that exact state if desired.

### DIFF
--- a/python/PiFinder/main.py
+++ b/python/PiFinder/main.py
@@ -18,6 +18,7 @@ import os
 import sys
 import logging
 import argparse
+import pickle
 from pathlib import Path
 from PIL import Image, ImageOps
 from multiprocessing import Process, Queue
@@ -167,7 +168,6 @@ def main(script_name=None, has_server=False):
 
     # Set path for test images
     root_dir = os.path.realpath(os.path.join(os.path.dirname(__file__), "..", ".."))
-    test_image_path = os.path.join(root_dir, "test_images")
 
     # init queues
     console_queue = Queue()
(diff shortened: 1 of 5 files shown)


python/PiFinder/main.py
```diff
@@ -498,29 +498,50 @@ def main(script_name=None, has_server=False):
...
                                 debug_image = subtract_background(debug_image)
                                 debug_image = debug_image.convert("RGB")
                                 debug_image = ImageOps.autocontrast(debug_image)
-                                debug_image.save(f"{test_image_path}/{uid}_sub.png")
+                                debug_image.save(
+                                    f"{utils.debug_dump_dir}/{uid}_sub.png"
+                                )
+
+                                # Screenshot
+                                ss = current_module.screen.copy()
```
mrosseel, line 512: why not do this before the writing, seems better to first gather all artifacts and then write them

mrosseel (approved):

python/PiFinder/main.py
brickbots: Good point, I'll move this up near the other bits!

PR #88, opened 2023-09-11
1. better handling of non-existing observation db

mrosseel:
- when creating the ObservationDatabase it checks whether the file exists, if not the file and tables are created
- includes crash fix for the server
- re-adds the screen into the shared_state


brickbots:
Fantastic!  Nice to move the creation logic into the class and remove it from the various places the database might be accessed.  Thank you!!

PR #90, opened 2023-09-16
Use icons as indicators for gps/camera/imu

mrosseel:
- camera debug now uses the configured exposure time so realistic plate solve times can be simulated
- when moving the unit, a moving icon is shown
- when a camera solve is done, the camera icon is shown
- gps icon is shown on gps_lock (shouldn't this include time lock as well???)
- while moving, show a greyed out constellation name when using IMU solves

Please test on your unit and let me know if anything needs tweaking.


python/PiFinder/solver.py
```diff
@@ -62,7 +62,7 @@ def solver(shared_state, solver_queue, camera_image, console_queue):
                     else:
                         solved["imu_pos"] = None
                     solved["solve_time"] = time.time()
-                    solved["cam_solve_time"] = time.time()
+                    solved["cam_solve_time"] = solved["solve_time"]
```
brickbots, line 65: 👍

python/PiFinder/calc_utils.py
```diff
@@ -53,3 +53,12 @@ def radec_to_altaz(self, ra, dec, alt_only=False):
         else:
             az = 360 - _az
         return alt, az
+
+
+import json
```
brickbots, line 58: I'd like to keep all module level imports up top.  After merging I'll re-arrange these :-)

python/PiFinder/state.py
brickbots, line 1: All great stuff here, thanks for making the SharedStateObj more friendly!

brickbots (approved):

PR #92, opened 2023-09-19
Adding a screen saver

mrosseel:
- added now config option 'Screen Off'
- When the unit is idle for this amount of time, the screen is turned off, preventing burn-in and lowering power consumption.
- when pressing a key to get out of screen off/lower brightness mode, the key is now properly ignored

See the .hide() method in the docs below:
https://luma-oled.readthedocs.io/en/latest/api-documentation.html#module-luma.oled.device

### DIFF
--- a/default_config.json
+++ b/default_config.json
@@ -13,6 +13,7 @@
     "screen_direction": "right",
     "solver_debug": 0,
     "sleep_timeout": "30s",
+    "screen_off_timeout": "Off",
     "solve_pixel": [256, 256],
     "catalogs": [
         "NGC",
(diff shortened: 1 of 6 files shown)


mrosseel:
@brickbots for screen dimming the Off setting is 100000 secs or 27 hours. Is this a feature or can I replace this with -1 and handle it in code to really mean Off ? That's how I did it now let me know if I should revert it. I wouldn't mind a failsafe where the screen turns off after a few hours, but not sure about it.

brickbots (approved):
Looks great!  Thanks for fixing the comment typo's along the way :-)

brickbots:
The 100000 seconds was just to have some value to avoid checking for the 'off' condition.  Your implementation is much more legible and future proof 👍 

The min/max for the keypad PWM is 0/100.  Basically the percent of the duty cycle the output is high.

PR #99, opened 2023-09-29
More catalogs + Supporting functionality

brickbots:
Working to add a couple more catalogs, but wanted to make the growing number of catalogs more usable

- [x] Ability to choose active catalogs
- [x] Limit visible catalogs to first 10 in config
- [x] Switch config.json to data dir
- [x] Add bright star catalog
- [x] Add Herschel 400


brickbots:
@mrosseel I know we are closing in on an RC, but I really wanted to make good on a couple of catalogs that people have been after.  With the ever expanding number, I thought I'd put some quick usability updates in:

* config.json moved to user data dir (makes re-ordering/selecting visible catalogs easier)
* Only the first 10 catalogs in the config are visible at all
* Catalogs can be selected as 'Active' (show up when cycling) via Catalog options page
* Added bright stars, and will add Herschel tomorrow

astro_data/bright_stars.csv
```diff
@@ -0,0 +1,79 @@
+Nr,NamedStar,LatinDesignation,RA Hr,RA Min,Dec Deg,Dec Min,Magnitude,Const
+0,Acamar,Theta Eridanus,2,58.2,-40,18,3.2,Eri
```
mrosseel, line 2: the new font can use real greek letters, which is neat to see but this is probably more user friendly

astro_data/herschel400.tsv
mrosseel, line 1: it would be interesting to note the origin of this file somewhere, e.g. in the setup.py

mrosseel (approved):
utACK, great upgrade!

astro_data/herschel400.tsv
brickbots: I wish I had good provenance for it!  It's from an old spreadsheet I had, which I scraped from a web page table at some point.   I'm pretty sure it's from the Saguaro astro club, but they changed their website and I can't be sure.   The descriptions seem to match what's in their master DB, so I'll add a comment!

PR #102, opened 2023-09-30
UI tweaks

brickbots:
- Rename 'Preview' to 'Camera'
- Add 'Camera' to main screen rotation (just 'A')
- Add option to adjust amount of BG sub in 'Camera' with default at 50%
  - This is mainly to make it clear that this is different than Chart and that it's showing a live view
  - Also avoids the completely black overexposed situation
- Change 'B' button on catalog to switch between description / image
- Add ** Logged x times at bottom of object descriptions
- Function button hints, with customizable timeout


mrosseel (approved):
utACK

Issue #100, opened 2023-09-29
Allow usernames other than 'pifinder'


hovissimo:
The install script hard-codes some paths for the `pifinder` username.

There is a conversation in Discord here: https://discord.com/channels/1087556380724052059/1087556507933089874/1157420010860445737

hovissimo:
It seems that I can't add a branch to open a PR, but here's the commit I would submit for this issue.


```
Author: Hovis <github@bunkercode.com>
Date:   Fri Sep 29 14:15:58 2023 -0700

    Use $HOME in pifinder_setup.sh

    This abstracts the username and prevents errors when users pick a
    username when using the Raspberry Pi imager.

    Resolves #100

diff --git a/pifinder_setup.sh b/pifinder_setup.sh
index a693a56..cb14fe2 100644
--- a/pifinder_setup.sh
+++ b/pifinder_setup.sh
@@ -6,27 +6,27 @@ cd PiFinder
 sudo pip install -r requirements.txt

 # data dirs
-mkdir ~/PiFinder_data
-mkdir ~/PiFinder_data/captures
-mkdir ~/PiFinder_data/obslists
-mkdir ~/PiFinder_data/screenshots
-mkdir ~/PiFinder_data/solver_debug_dumps
-mkdir ~/PiFinder_data/logs
-chmod -R 777 ~/PiFinder_data
+mkdir $HOME/PiFinder_data
+mkdir $HOME/PiFinder_data/captures
+mkdir $HOME/PiFinder_data/obslists
+mkdir $HOME/PiFinder_data/screenshots
+mkdir $HOME/PiFinder_data/solver_debug_dumps
+mkdir $HOME/PiFinder_data/logs
+chmod -R 777 $HOME/PiFinder_data

 # Wifi config
-sudo cp ~/PiFinder/pi_config_files/dhcpcd.* /etc
-sudo cp ~/PiFinder/pi_config_files/dhcpcd.conf.sta /etc/dhcpcd.conf
-sudo cp ~/PiFinder/pi_config_files/dnsmasq.conf /etc/dnsmasq.conf
-sudo cp ~/PiFinder/pi_config_files/hostapd.conf /etc/hostapd/hostapd.conf
-echo -n "Cli" > ~/PiFinder/wifi_status.txt
+sudo cp $HOME/PiFinder/pi_config_files/dhcpcd.* /etc
[...]

PR #104, opened 2023-10-04
Allow arbitrary username

hovissimo:
This PR removes the requirement that the user select 'pifinder' as the username for their Raspberry pi.

Overview of changes:
 - updates hard-coded user dir paths with $HOME
 - moves pifinder.service to pifinder.service.template and updates the install script to interpolate that template back to pifinder.service with the current user's username
 - One incidental change updating `astro_data_dir` to a consistent reference to utils

I tested these changes by installing PiFinder on a new sd card with my own username.

### DIFF
--- a/pi_config_files/smb.conf
+++ b/pi_config_files/smb.conf
@@ -235,7 +235,7 @@ guest account = pifinder
 # to the drivers directory for these users to have write rights in it
 ;   write list = root, @lpadmin
 [shared]
-path=/home/pifinder/PiFinder_data
+path=$HOME/PiFinder_data
 writeable=Yes
 create mask=0777
 directory mask=0777
--- a/docs/dev_guide.md
+++ b/docs/dev_guide.md
@@ -25,7 +25,7 @@ pip install -r requirements_dev.txt
 
 Download the hipparcos catalog, change the download location.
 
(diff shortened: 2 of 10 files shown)


hovissimo:
This resolves https://github.com/brickbots/PiFinder/issues/100

pi_config_files/smb.conf
```diff
@@ -235,7 +235,7 @@ guest account = pifinder
 # to the drivers directory for these users to have write rights in it
 ;   write list = root, @lpadmin
 [shared]
-path=/home/pifinder/PiFinder_data
+path=$HOME/PiFinder_data
```
brickbots, line 238: I'm not sure $HOME will work in this context as the samba service is not running as the `pifinder` user.  Perhaps the same trick you used to inject the current user into the service definition file would work here?

brickbots (commented):
Thank you for this :+1:  Sorry it's taken me a while to get to this PR, but I'd like to get it merged as part of an upcoming point release.   It all looks really good, but I had a question about the samba share config which I noted in another comment.

brickbots:
Thank you @hovissimo for the contribution, it will come in useful in the future.  I'm closing this particular PR as we are starting to move to a new OS release (bookworm) and some of this will need to be re-worked a bit.

PR #113, opened 2023-11-13
1. Skysafari pushto

mrosseel:
GoTo on skysafari now sends the coords to the pifinder, which shows it in a locate screen.

- Created a UiState which is now in the shared_state, because the pos_server runs in another thread and needs to communicate the new target position to the rest of the application
- Added necessary handling of the goto message from skysafari to pos_server
- refactored Catalogs for easier handling of pushto target and future developments

Known issue:
- there's a crash in the push cat / push near objects methods, fixed in the planets PR (messy to backport)


brickbots (approved):
Okay!  There is A LOT here, thanks for all the hard work 🥇   I very much appreciate your continued dedication to type hinting, I'm still planning on adding mypy checking to our workflow at some point and it's nice to have some of the code base up to snuff.

There are some things here (history vs. observing list) that will eventually phase out, but solidifying the ui state into an object and moving it to the shared state is good to have done.

I think we should merge this as soon as the bookworm stuff is ready.  I don't think anything I'm doing there will conflict, so it should be easy to get all these branches merged into main when ready!

PR #120, opened 2023-12-19
1 - add fake sys_utils

mrosseel:
Quick fix for all non-raspberry pi computers wanting to run pifinder.


brickbots (approved):
This works for now!  I'm thinking the longer term play here might be to make everything in this module one 'system' class and we create a class factory that returns the right one depending on some system check.  That would avoid having to do the try/except import in multiple places.

Just leaving this here for future work!

PR #121, opened 2023-12-20
2 - Add advanced view in the web UI for development

mrosseel:
this is a quick fix to expose the gps-lock and time-lock endpoints via the UI. This is helpful for debugging when no GPS lock is present (both on desktop and in-house pifinder which can't find a lock)

Note:
The good solution is to add a little pencil next to the gps lock entry on the index, leading to a /gps page where you can enter lat/lon and time. Time could be gotten from the browser, lat/lon maybe also

### DIFF
--- a/python/PiFinder/server.py
+++ b/python/PiFinder/server.py
@@ -24,7 +24,7 @@ from PiFinder.keyboard_interface import KeyboardInterface
 try:
     from PiFinder import sys_utils
 except ImportError:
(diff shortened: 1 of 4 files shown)


brickbots:
Works nice as a temp fix.  As you commented, the right way is to add an edit icon on the first page and pull the time/location from the phone/computer.  It's on my todo list, but just a ways down.  This is nice to have in the meantime 👍

PR #123, opened 2023-12-30
2. Planets

mrosseel:
Adding a planet catalog.

Future todo's:
- show Planet names instead of PL 1 ...
- Magnitudes for sun/moon


.pre-commit-config.yaml
```diff
@@ -2,7 +2,7 @@
     # See https://pre-commit.com/hooks.html for more hooks
     repos:
       - repo: https://github.com/ambv/black
-        rev: 23.3.0
+        rev: 23.12.1
         hooks:
           - id: black
-            language_version: python3.9
+            language_version: python3.12
```
brickbots, line 8: Is it okay if we settle on 3.11 for the bookworm release?  Not sure what new stuff is in 3.12, but we'd have to roll our own for the Raspberry Pi OS

brickbots (commented):
This looks pretty good as well, with the 3.11 note.  A bit hard to suss out just the planet stuff as it's got the other object push bits in there, but I think it's good so far!

.pre-commit-config.yaml
mrosseel: yes, 3.11 should be fine

mrosseel:
> This looks pretty good as well, with the 3.11 note. A bit hard to suss out just the planet stuff as it's got the other object push bits in there, but I think it's good so far!

once the skysafari stuff is merged the diff will make more sense. I could base it on the other PR but not sure how github behaves in that case with merging etc.

.pre-commit-config.yaml
mrosseel: rolled this back

PR #124, opened 2024-01-17
Make latitude, longitude, altitude and time settable via the web UI

mrosseel:
This helps in case of either forgetting your GPS dongle, or having a malfunctioning internal GPS.

Possible future work:
- it might be nice to have preset locations like HOME, Observatory, ... similar to how the different wifi networks are configured.

### DIFF
--- a/python/PiFinder/server.py
+++ b/python/PiFinder/server.py
@@ -1,9 +1,9 @@
-import time
 import logging
 import io
 import uuid
 import json
 from datetime import datetime, timezone
+import time
 from bottle import (
     Bottle,
     run,
@@ -46,12 +46,21 @@ def auth_required(func):
 
 
 class Server:
-    def __init__(self, q, gps_queue, shared_state):
+    def __init__(self, q, gps_queue, shared_state, debug=False):
         self.version_txt = f"{utils.pifinder_dir}/version.txt"
         self.q = q
         self.gps_queue = gps_queue
         self.shared_state = shared_state
         self.ki = KeyboardInterface()
+        # gps info
+        self.lat = None
+        self.lon = None
+        self.altitude = None
+        self.gps_locked = False
+
+        logger = logging.getLogger()
+        if debug:
+            logger.setLevel(logging.DEBUG)
 
         button_dict = {
             "UP": self.ki.UP,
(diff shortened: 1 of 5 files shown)


python/views/header.tpl
```diff
@@ -25,6 +26,7 @@
         <li><a href="/">Home</a></li>
         <li><a href="/remote">Remote</a></li>
         <li><a href="/network">Network Setup</a></li>
+        <li><a href="/gps">GPS Setup</a></li>
```
brickbots, line 29: I'd rather have this on the 'Tools' page as it should (ideally) be rarely used.  I know it's come up a lot in the discord (people not using GPS, issues with locks), but I hope these are not a representative sample of users. If people are using this often I have failed in some significant way as it's absolutely NOT the intended (or a good) user experience.

brickbots (commented):
Thank you so much for working to add this!  I'm excited to actually try this and see how it works in context.

python/PiFinder/server.py
```diff
@@ -356,24 +398,24 @@ def serve_pil_image():
 
         @app.route("/gps-lock")
```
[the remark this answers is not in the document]
mrosseel: they used to be called with the old web interface, there were dedicated buttons for testing. Now it's probably no longer needed, I'll see if I can remove it.

python/views/header.tpl
mrosseel: then I'd need to integrate it with the Tools page, and it's a lot of code, tools might become a garbage bin then. Maybe not adding it to the menu is enough? On save it returns to home

brickbots: Yes... tools is probably not the right place.  How about not on the menu, but accessible from the main page via the edit icon?

mrosseel: that's what I did in the last commit

python/views/index.tpl
```diff
@@ -18,6 +18,7 @@
       <i class="material-icons medium">{{gps_icon}}</i>
     </td>
     <td class="grey-text text-lighten-1">{{gps_text}}<br>lat: {{lat_text}} / lon: {{lon_text}}</td>
+    <td><a href="/gps" class="grey-text"><i class="material-icons">edit</i></a></td>
```
brickbots, line 21: Sorry @mrosseel I missed this edit to the index page!  Merging now!

PR #127, opened 2024-01-23
Nearby mode

mrosseel:
Up till now, you kinda had to know what you were looking for before you could 'pifind' it.

Not anymore, with the addition of 'browsing mode'. Alternative names could be 'explore mode' or 'tourist mode' - TBD

With browsing mode you can put your pifinder on the browsing screen, point your telescope at any part of the sky and the screen will show the 10 closest objects.

This allows the user to quickly go to the interesting stuff in the neighborhood (kind of like the push near feature but on steroids). If you know your deepsky, you could even put the scope in the general direction of an object, spot it in the list and finetune its location.

Screenshot 2024-01-23 at 22 01 16

TODO:
- honor the filter settings of the catalog screen
- support up/down to select an entry and either expand it or allow selection and switch to the locate/detail screen.


brickbots (approved):
Fantastic!  I'm looking forward to testing it 💯

PR #134, opened 2024-01-30
Cedar solve

mrosseel:


requirements.txt
```diff
@@ -4,7 +4,7 @@ bottle==0.12.25
 cheroot==10.0.0
 gpsdclient==1.3.2
 luma.oled==3.12.0
-pillow==9.5.0
+pillow
```
brickbots, line 7: Was 9.5.0 not working in some way?  If not, what version did pip end up pulling for you so we can add it here (`pip list`)?  I like to keep the requirements file version locked just to avoid any mysterious issues from one image build to the next.

requirements.txt
```diff
@@ -17,4 +17,5 @@ sh==1.14.3
 skyfield==1.45
 timezonefinder==6.1.9
 tqdm==4.65.0
-alpyca
+grpcio
```
brickbots, line 20: Let's also specify the versions here :+1:

brickbots (approved):
This is fantastic and I'm excited to test it.  I left a couple of small notes/questions.  I'm happy to merge and then make the minor changes if you want to call this done.  Just let me know what you'd like to do 👍

PR #140, opened 2024-02-19
Fix Issue 130

mrosseel:


brickbots (approved):
This is a nice QOL fix!  We could tidy up the catalog source data, but some will probably still slip through so this is a nice backup to have :+1:

PR #144, opened 2024-02-21
Fixed selection bugs in catalogs. Fix nearby crash.

mrosseel:
also added pre-commit to nix shell


brickbots (approved):
Thanks for patching this up and pushing it through!

Maybe the catalog system is due for another refactor now that we have added heaps more functionality?  A bit of tech-debt for another day I think :-)

mrosseel:
> Thanks for patching this up and pushing it through!
> 
> Maybe the catalog system is due for another refactor now that we have added heaps more functionality? A bit of tech-debt for another day I think :-)

The main classes are mostly fine, but the periphery can use some work like the CatalogDesignator and the CatalogTracker. If you have ideas let's discuss

PR #148, opened 2024-03-03
Remove catalog config limit

mrosseel:
We now have a scrolling menu for catalog selection, so the nr of catalogs can be unlimited !


brickbots:
This seems like it will do the trick!  Very quick work.  I'm looking forward to testing it live!

PR #150, opened 2024-03-05
Solver bug

mrosseel:
- When no centroids are detected, we now continue instead of return
- max-size is now 10 instead of 8, because on initial alignment the brightest stars are not detected, and these are the ones we typically align on
- add an empty debug camera image so we switch between solving and and not solving (this fails without the return fix)


python/PiFinder/camera_debug.py
```diff
@@ -40,7 +42,8 @@ def capture(self) -> Image.Image:
         sleep_time = self.exposure_time / 1000000
         time.sleep(sleep_time)
         logging.debug("CameraDebug exposed for %s seconds", sleep_time)
-        return self.image
+        self.image_bool = not self.image_bool
+        return self.image if self.image_bool else self.image2
```
brickbots, line 46: This is a very nice addition.  Eventually we might want to expand this even more to just cycle through all images in dir :+1:

brickbots (approved):
Thanks for this!  Do we want to undo my change to turn off the Cedar detect?

PR #164, opened 2024-05-02
Imu fliplock

brickbots:
When starting up the IMU can accumulate offsets before it reports fully calibrated.  In this case the seeded quad_history of 0,0,0,0 will always generate a large offset and be detected as a flip.  This merge request counts sequential flips and if there are more than 10, it seeds the history with the current readings to 'reset' the system


brickbots:
Hotfixing to Release

PR #165, opened 2024-05-03
Use filtered objects for nearby balltree

brickbots:
It seems that the Nearby feature was not using the filtered object list from the catalog object.  

* Add filtered arg for Catalogs.get_objects()
* Overload Catalog.get_objects() to look for filtered


python/PiFinder/catalogs.py
```diff
@@ -262,6 +262,9 @@ def filter_objects(self, shared_state) -> List[CompositeObject]:
         self.last_filtered = time.time()
         return self.filtered_objects
 
+    def get_filtered_objects(self):
```
mrosseel, line 265: the state of the self.filtered_objects starts out as all the objects, and only after filter_objects is called it is changed. So not sure why we shouldn't call filter_objects first before returnng self.filtered_objects. This needs the shared_state, mainly for altitude calculations I think

brickbots: I'm going to merge this, as it works mostly and is a big improvement... but the whole filtering system needs a bit of an overhaul as it still does not really take into account observed status through a session.  I'll open an issue for that.

PR #166, opened 2024-05-03
1. Fixing the missing names bug

mrosseel:


brickbots:
It will be great to have these back!

PR #171, opened 2024-05-13
Status screen shows nr of GPS sats seen/used

mrosseel:
Screenshot 2024-05-14 at 16 57 15


brickbots:
This is a great 'little' feature that I think will be very useful!  Thank you for taking this on :+1:

PR #184, opened 2024-07-07
Hard stop dovetail

jwlcurtis:
Added a version of the dovetail bottom with a hard stop to allow for a more secure mount.


brickbots:
Thank you so much for this great addition @jwlcurtis and I appreciate the updated fusion file as well 🎉   I believe there may need to a different versions of this part for left and right hand facing PiFinders.  Do you have time to make a second version with the hard-stop at the opposite end?

jwlcurtis:
@brickbots I created a second version with the hard-stop on the opposite end. It retains the circular cut-out at the end so there hopefully shouldn't be any issues during assembly.

brickbots:
Awesome, thank you for this and sorry for the delay in reviewing!

Issue #117, opened 2023-11-29
Chart rotation potentially incorrect very near celestial poles


brickbots:
When aligned on a pixel off-center from the camera image, the chart rotation appears incorrect when the celestial pole is between the geometric center of the camera image and the target pixel.

Tetra3 returns roll based on the geometric center, not the target pixel.  According to Gustav this would be tricky to implement.

A potential solution is to render the chart based on the geometric center of the camera image (raw solution values + roll) then offset  this render by the target pixel values to align it with the target pixel before copying it to the screen and adding the reticle.  This would ultimately center the chart on the target pixel, but provide a more stable rotation reference.

TakKanekoGit:
@brickbots -- I will try to pick this issue up in step 2 of the "Roll Issue Fix". 

As I understand from your comments above, the roll measured by plate solving is only valid at the centre of the camera FoV. The difference between the roll at the camera centre and the target pixel becomes bigger higher declinations. If we want to use the roll as a measurement, as proposed for later enhancements, I think we need to fix this?

Here's how I propose to fix it:

1. Take the (ra_camera, dec_camera, roll_camera) from plate solving. Compare the measured roll against the roll calculated at (ra_camera, dec_camera) against the measured roll_camera to get the roll_offset. This should be reasonably constant for an Alt/Az mount.
2. Work out the (ra_scope, dec_scope) at the target pixel as done currently. Calculate the roll at the target and add the roll_offset. This should give the correct roll as seen by the scope.

Do you think this will work or is there more to the problem?

PR #185, opened 2024-07-07
Roll calculation

TakKanekoGit:
First try at resolving the roll issue. During IMU dead-reckoning, integrator.py calls calc_utils.sf_utils.radec_to_roll() to calculate the expected roll at (RA, Dec) give the observer's location and time. The roll value is saved to solved["Roll"].

The functions/methods have been unit tested but not system tested in a PiFinder. It also needs to be field tested to check it fixes the roll issue.

### DIFF
--- a/python/PiFinder/calc_utils.py
+++ b/python/PiFinder/calc_utils.py
@@ -192,6 +192,56 @@ def calc_object_altitude(shared_state, obj) -> Optional[float]:
     return None
 
 
+def hadec_to_pa(ha, dec, lat):
+    """ 
+    Returns the parallactic angle of an object at (ha, dec) for an observer
+    at latitude lat. 
+    
+    Note that all angles are in radians.
+
+    The parallactic angle is the angle between the great circles between the 
+    zenith (Z) to the source (S) and the North Pole (P) to the source. By 
+    convention, the parallactic angle is measured from PS to ZS, positive 
+    towards East. The parallactic angle is negative when H < 0 and positive 
+    when H > 0. When At the meridian (i.e. H=0), the parallactic angle is 0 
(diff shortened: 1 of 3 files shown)


brickbots:
Thank you @TakKanekoGit for all this great work and for including great tests along with it!  I think this is ready to merge and test under the stars :+1:

TakKanekoGit:
Thank you for sorting out the linting and the tip about Black. I'll check it before I contribute code next time. Out of interest, do you usually run it at the end or use a Black extension in your editor?

Hope it works under the skies. Fingers crossed.

brickbots:
My pleasure!  I'm very excited to test this out and I appreciate the great comments and overall code quality 💯 

I use Black as a pre-commit hook as I develop a fair bit on the actual PiFinder via SSH/tmux/Vim, but I am pretty sure it can be integrated in various editors.

In the very near future we'll be moving to a more robust set of tools orchestrated through nox:
Ruff - Faster linter/formatter which implements the same basic rules as Black
MyPy - Static type analysis, we have a branch with appropriate code changes, rules/ignores so that the current code base passes
PyTest - Switching from unittest to PyTest.  I've already ported your great calc_utils test over to PyTest in this branch

So if you are going to be setting up a dev environment, this would probably be a better target as in the next few weeks the project will switch over 👍

TakKanekoGit:
Thank you for the heads on the upcoming new tools. Very interested to try them out when the switch over happens.

I'll work on step 2 of the roll issue fix (taking into account the offset) after the summer break and test it on the PiFinder when all the bits arrive. As part of step 2, I will also try to fix issue #117. I'll add the proposed solution direction in the issue to check with you.

https://github.com/brickbots/PiFinder/issues/117

Thanks for your help!

PR #187, opened 2024-07-14
Add T9 style keypad for searching

mrosseel:


python/PiFinder/db/objects_db.py
jscheidtmann, line 1: In memory you‘re using `*.lower()`, I would expect a similar construct for accessing DB.

python/PiFinder/main.py
jscheidtmann, line 1: Why do you comment out GPS debug messages, here?

mrosseel: I added them some time ago when debugging gps issues, but they are very noisy and overwhelm all other debugging, so I removed them. Better would be to be able to switch on/off debugging for certain modules, but that's for another PR.

python/PiFinder/db/objects_db.py
mrosseel: true, althought the DB version is not used atm. I'll update it to also use lower.

brickbots:
Thank you @mrosseel for this AWESOME new feature 🎉 

@jscheidtmann I appreciate you checking over this PR and adding your questions/thoughts!

PR #190, opened 2024-07-15
Object list scrolling text with common names

mrosseel:
Remaining TODO's:
- where can I get the scrollspeed config from?

### DIFF
--- a/python/PiFinder/ui/object_list.py
+++ b/python/PiFinder/ui/object_list.py
@@ -22,6 +22,7 @@ from PiFinder.calc_utils import aim_degrees
 from PiFinder.catalog_utils import ClosestObjectsFinder
 from PiFinder import utils
 from PiFinder.catalogs import CompositeObject
+from PiFinder.ui.ui_utils import name_deduplicate
 
 
 class DisplayModes(Enum):
@@ -55,6 +56,7 @@ class UIObjectList(UITextMenu):
     bulb = "󰛨"
     star = ""
     ruler = ""
+
 
     def __init__(self, *args, **kwargs) -> None:
         # hack at our item definition here to allow re-use of UITextMenu
@@ -185,13 +187,14 @@ class UIObjectList(UITextMenu):
             return int(255 + ((125 - 255) / (16 - 9)) * (mag - 9))
 
     def create_name_text(self, obj: CompositeObject) -> str:
-        """
-        Returns the catalog code + sequence
-        padded out to be 7 chars
-        NGC0000
-        """
+        dedups = name_deduplicate(obj.names, [f"{obj.catalog_code}{obj.sequence}"])
+        result = ", ".join(dedups)
+        return result
+
+    def create_shortname_text(self, obj: CompositeObject) -> str:
         name = f"{obj.catalog_code}{obj.sequence}"
-        return f"{name: <7}"
+        return name
+        # return f"{name: <7}"
 
     def create_locate_text(self, obj: CompositeObject) -> str:
         az, alt = aim_degrees(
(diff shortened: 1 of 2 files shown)


python/PiFinder/ui/object_list.py
```diff
@@ -299,33 +335,55 @@ def update(self, force=False):
...
+                line_bg = 32 if is_focus else 0
                 marker = self.get_marker(_menu_item.obj_type, line_color, line_bg)
                 if marker is not None:
                     self.screen.paste(marker, (0, line_pos + 2))
 
+                # calculate start of both pieces of text
+                px_per_char = self.base_px_per_char if not is_focus else self.bold_px_per_char
+                chars_per_line = math.floor(self.display.width/px_per_char)
+                begin_x = 12
+                begin_x2 = begin_x + (len(item_name))*px_per_char
```
brickbots, line 357: Here, rather than `px_per_char` you can use `line_font.width` since line_font will be an instance of the PiFinder font class.

python/PiFinder/ui/object_list.py
```diff
@@ -299,33 +335,55 @@ def update(self, force=False):
...
+
+                # draw first text
                 self.draw.text(
-                    (12, line_pos),
-                    item_line,
+                    (begin_x, line_pos),
```
brickbots, line 361: Thank you for pushing this towards resolution independence 🙌

python/PiFinder/ui/ui_utils.py
```diff
@@ -4,6 +4,20 @@
 import math
 
 
+def calculate_fixed_width(font, pixel_width):
```
brickbots, line 7: These can both be removed in favor of the methods on the Font class

python/PiFinder/ui/object_list.py
```diff
@@ -236,6 +260,16 @@ def _obj_to_mag_color(self, obj: CompositeObject) -> int:
...
         return self._interpolate_color(self._safe_obj_mag(obj))
 
+    def _get_scrollspeed_config(self):
+        scroll_dict = {
+            "Off": 0,
+            "Fast": TextLayouterScroll.FAST,
+            "Med": TextLayouterScroll.MEDIUM,
+            "Slow": TextLayouterScroll.SLOW,
+        }
+        scrollspeed = self._config_options["Scrolling"]["value"]
```
brickbots, line 270: None of the UI modules will have _config_options any longer.  This can/should go into the main config file via the config object `self.config_object` which is a shared instance of the `Config` class which has these methods:

```python
    def set_option(self, option, value):
        self._config_dict[option] = value
        self.dump_config()

    def get_option(self, option, default: Any = None):
        return self._config_dict.get(
            option, self._default_config_dict.get(option, default)
        )
```

To allow people to configure this, we'll add a sub-menu to the menu configuration dictionary under 'Settings'... which I still need to finish porting some previous options to and probably add some more hierarchy to like a Display sub-menu!

brickbots (changes requested):
Thank you, I'm excited to try this on the PiFinder hardware!

python/PiFinder/ui/object_list.py
```diff
@@ -244,7 +278,8 @@ def active(self):
 
     def update(self, force=False):
         # clear screen
-        self.draw.rectangle([0, 0, 128, 128], fill=self.colors.get(0))
+        self.draw.rectangle((0, 0, self.display.width, self.display.height),
```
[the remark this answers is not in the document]
mrosseel: I didn't change this, just fixed the tuple vs list error - but I'll replace it with the clear_screen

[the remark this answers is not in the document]
brickbots: Thank you!  I should have replaced this when I was morphing the Nearby screen to the object_list and I appreciate you cleaning up my mess here :-)

python/PiFinder/ui/object_list.py
brickbots: Resolving as this is on my Todo list now 🤝

PR #192, opened 2024-07-15
Compensate for roll offset

TakKanekoGit:
Compensates for the roll offset between the roll seen by the camera and the expected roll at the (RA, Dec). This should also fix issue #117.

The roll at the camera center is different from the roll at the target pixel. To address this, the (RA, Dec, Roll) at the camera center get stored as solution["RA_camera"], solution["Dec_camera"] and solution["Roll_camera"]. The coordinates at the target pixels are still the same: solution["RA"], etc. 

The roll offset is calculated at the camera center in integrator.py and stored as solution["Roll_offset"] every plate solve. solution["Roll"] at the target is calculated from the expected roll at the target (RA, Dec) and compensated with the Roll_offset.

This should display the constellation with the correct roll as seen by an observer. This PR hasn't been system-tested on a PiFinder or field-tested. It's possible that the sign for the roll_offset is the wrong way round. If so, this can be changed in integrator.estimate_roll_offset().

### DIFF
--- a/python/PiFinder/calc_utils.py
+++ b/python/PiFinder/calc_utils.py
@@ -48,7 +48,9 @@ class FastAltAz:
             self.lat * math.pi / 180
         ) + math.cos(dec * math.pi / 180) * math.cos(
             self.lat * math.pi / 180
-        ) * math.cos(hour_angle * math.pi / 180)
+        ) * math.cos(
+            hour_angle * math.pi / 180
+        )
 
         alt = math.asin(_alt) * 180 / math.pi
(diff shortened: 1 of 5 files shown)


brickbots:
Thank you for this!  It looks great and I'll aim to get it tested under the stars and validated with a sample historical data set soon so it can be fully merged!

TakKanekoGit:
Closing this PR because of conflicts rebasing. Replaced by new PR "Compensate for roll offset (rebased) #192"

PR #194, opened 2024-07-18
Object_list improvements for Nearby mode

mrosseel:
- massive speedup of nearby code
- removal of old nearby ui file
- new nearby.py file containing all nearby-related code

Only remaining open points are the showing of sort order on the first line and the nearby-unique positioning of the list - we'll have to discuss


python/PiFinder/utils.py
```diff
@@ -57,3 +57,27 @@ def get_os_info():
     else:
         os_detail = "N/A"
     return os_detail, platform_system, architecture
+
+
+class Timer:
```
brickbots, line 62: Nice and handy!  Thank you for this!

python/PiFinder/ui/base.py
```diff
@@ -118,15 +118,15 @@ def clear_screen(self):
...
+    def message(self, message, timeout: float = 2, size=[5, 44, 123, 84]):
         """
         Creates a box with text in the center of the screen.
         Waits timeout in seconds
         """
 
         # shadow
         self.draw.rectangle(
-            [size[0] + 5, size[1] + 5, size[2] + 5, size[3] + 5],
+            (size[0] + 5, size[1] + 5, size[2] + 5, size[3] + 5),
```
brickbots, line 129: 😓 I'll be better about making these bounding-box coordinates tuples

brickbots (approved):
Great, I'm merging this in!

PR #196, opened 2024-07-20
First pass at marking menus

brickbots:
Add marking menu system
* Long Square launches menu
* Arrows select option in that direction
* Calls marking_menu_[direction] function on UIModule class when selected
* Check ui/object_list.py and ui/base.py for example implementation

Object list:
Screenshot 2024-07-20 at 8 33 22 AM

Marking menu with three enabled options (down, left, right)
Screenshot 2024-07-20 at 8 33 36 AM


python/PiFinder/ui/object_list.py
```diff
@@ -106,6 +113,13 @@ def __init__(self, *args, **kwargs) -> None:
         self.sort_cycle = cycle(SortOrder)
         self.current_sort = next(self.sort_cycle)
 
+        if self.current_sort == SortOrder.CATALOG_SEQUENCE:
```
mrosseel, line 116: shouldn't the marking menu be a class where you can call a method to enable/disable items?

python/PiFinder/ui/object_list.py
```diff
@@ -477,11 +506,22 @@ def key_square(self):
             self.current_mode = next(self.mode_cycle)
             self.refresh()
 
-    def key_plus(self):
+    def marking_menu_left(self):
```
mrosseel, line 509: maybe one method to implement with the option and the level as params?

python/PiFinder/ui/object_list.py
brickbots: Yeah, agreed.  The developer ergonomics here are not great.  I'll rework this with the multi-level functionality as the API will need to change and I'll think a bit more about the ergonomics.

python/PiFinder/ui/object_list.py
brickbots: I was debating allowing arbitrary callable here as the API, rather than distinct callbacks.  I think with the switch to multi-level, it will need to be done with callable passed as part of the marking_menu configuration 👍

PR #197, opened 2024-07-21
UI buttons for website

mrosseel:
update the remote page of the webserver to show the correct keys


brickbots:
Thank you for spotting this and updating 🥇

PR #198, opened 2024-07-22
1. comets

mrosseel:
TODO:
- [x] show some info on the age of the elements
- [x] does it work offline, how do we force an update? ==> retries every minute now
- [x] how accurate are mags?
- [ ] sort on mags so we can have the brightest comets on top ==> should be done in new PR
- [x] measure performance or hide the cost (now it's at startup and takes a few seconds)
- [x] update positions every minute


brickbots:
Okay, lots here but looks good.  Will need to be tested in context.  Thanks for all the work on this @mrosseel !

PR #201, opened 2024-07-29
Always run centroider, even if cedar_detect is not found

jscheidtmann:
When running on another architecture, the cedar_detect binary might not be available. In this case always run the tretra3 centroider.


jscheidtmann:
Sorry, based this off of the wrong branch.

PR #202, opened 2024-07-29
Always run centroider, even if cedar_detect is not found

jscheidtmann:
When running on another architecture than the Pi, the cedar_detect binary might not be available. In this case always run the tretra3 centroider.

### DIFF
--- a/python/PiFinder/solver.py
+++ b/python/PiFinder/solver.py
@@ -14,14 +14,13 @@ import logging
 import sys
 from time import perf_counter as precision_timestamp
 
-logger = logging.getLogger("solver")
-
 from PiFinder import utils
 
 sys.path.append(str(utils.tetra3_dir))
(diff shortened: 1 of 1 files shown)


mrosseel (commented):
it's good that your code doesn't just assume the binaries are there. I'm not so sure that we should fallback - I know it will solve your issues on windows, but maybe getting the windows binary would be the best solution here? 
My opinion would be: let's go ahead with this, but try to get the windows binary in at some point, and then revisit this to detect missing binaries and exit, because I think it should never be run like this.

brickbots:
Thanks @jscheidtmann for working to make development easier on Windows!  I agree with @mrosseel that we should strive to get Cedar working on Windows to help make the various execution environments as close as possible, but this is a great improvement to the current state of play 👍

PR #203, opened 2024-07-29
Harmonize logging and make it work(tm) with multiprocessing

jscheidtmann:
Most of the logging code so far uses the root logger by using `logging.debug()` and friends. Configuration of the log levels and handler is done inline as code. This way, it is impossible to store and use logging configurations for different purposes (e.g. debug a specific device, or interprocess communication or ..) nor switch between these configurations. 

In addition the [logging cookbook](https://docs.python.org/dev/howto/logging-cookbook.html#logging-to-a-single-file-from-multiple-processes) tells us: 

 > logging to a single file from multiple processes is not supported, because there is no standard way to serialize access to a single file across multiple processes in Python.

This pull request therefore aims at: 

 - [x] Using `logging.getLogger("MeaningfulName")` in all relevant compile units 
 - [x] Convert formatted strings to `printf` semantics, as supported by logging functions, which avoids rendering the string, if the log level is not enabled.
 - [x] Introduce a separate process to consolidate logs, as recommended by the logging cookbook (see link above)
 - [x] Read logging configuration from a file during start-up.
 - ~~Provide some example logging configurations, that provide insights into different parts of PiFinder.~~
 - [x] Provide a logging configuration showing all logging options. 
 
 Note that only operating systems will be supported in the first go, that obey fork semantics, that means where the logging configuration is inherited from the main process (i.e. not on Windows).


mrosseel:
hi, great initiative, it has bothered me as well and I've done some experiments but never committed to go all the way.
Multiprocessing logging indeed doesn't work reliably currently, and makes debugging harder than it should be.

jscheidtmann:
@brickbots @mrosseel Could you please review and merge? I will be out on vacation starting Monday for 2 weeks.

.github/workflows/nox.yml
```diff
@@ -11,4 +11,4 @@ jobs:
       - uses: wntrblm/nox@2024.04.15
         with:
           python-versions: "3.9"
-      - run: nox -s lint format type_hints smoke_tests unit_tests
+      - run: nox --force-python 3.9 -s lint format type_hints smoke_tests unit_tests
```
brickbots, line 14: There was an oversight in the noxfile that allowed some of the tests to run on arbitrary python versions.  This has been fixed so I'll revert this after merge :+1:

python/requirements.txt
```diff
@@ -4,6 +4,7 @@ bottle==0.12.25
 cheroot==10.0.0
 gpsdclient==1.3.2
 grpcio==1.60.0
+json5
```
brickbots, line 7: Not that requirements are added to frequently, but for potential future reference, I prefer to have specific versions listed here. I'll update this after merging :+1:

python/tests/test_multiproclogging.py
```diff
@@ -0,0 +1,157 @@
+import pytest
```
brickbots, line 1: Thank you SO much for adding tests 🫶

brickbots (approved):
This is an awesome improvement @jscheidtmann and very much appreciated.  Very nice work, I appreciate the non-related bits of improvement along the way like the comment clarification and better log messages.

I've dropped couple of comments for reference, but I'll take care of these very minor changes after merge.  Thank you! 💯

PR #206, opened 2024-08-10
Fix solver logging

jscheidtmann:
Seems one commit was not part of the previous pull request: Maybe I reverted it inadvertently during merging.


brickbots:
Merging!  Thank you for the quick followup @jscheidtmann and for the original contribution 🙌

PR #207, opened 2024-08-10
[WIP] Recent/History view

mrosseel:
An object view with a list of objects that have been looked at since startup, ordered from most recent to least recent.


python/PiFinder/composite_object.py
```diff
@@ -2,10 +2,20 @@
 from dataclasses import dataclass, field
 import numpy as np
 import json
-from PiFinder.utils import is_number
 from typing import List
 
 
+def is_number(s):
```
brickbots, line 8: I'm not seeing where this is used, but it seems like the same number of lines could be used to just ask forgiveness rather than permission inline
```
if is_number(x):
    f=float(x)
    # do the float thing with f
else:
    # do the non float thing
```
vs
```
try:
    f=float(x)
    # do the float thing
except (ValueError, TypeError):
    # do the non float thing
```

Using the function might be nice if there ends up being a lot of stuff going on in the 'float stuff here' section which could also throw a ValueError.  

If you do want to keep this function, no need for the s is None check if you catch both Value and Type errors :+1:

python/PiFinder/state.py
```diff
@@ -11,14 +11,34 @@
 import pytz
 from PiFinder import config
 import logging
+from typing import List
+from PiFinder.composite_object import CompositeObject
+
+
+class Recent:
```
brickbots, line 18: How about using `__iter__` and `append` for this and naming it RecentList?  The semantics would then be just like a python list 🙌   You could even sub-class list itself and replace just the `append` method to implement the same uniqueness/ordering your have here


```
class RecentObjectsList(list):
    def append(self, v: CompositeObject):
        try:
              # remove the object if it's already here, so we can add it to the end
              super().remove(v)
         except ValueError:
             pass
         # add it to the end of the list
         super().append(v)
```

python/PiFinder/state.py
```diff
@@ -31,11 +51,11 @@ def observing_list(self):
     def set_observing_list(self, v):
         self.__observing_list = v
 
-    def history_list(self):
-        return self.__history_list
+    def recent_list(self) -> List[CompositeObject]:
+        return self.__recent.recent_list()
```
brickbots, line 55: If Recent becomes RecentList then you can just return self.__recent here

python/PiFinder/ui/object_history.py
brickbots, line 1: Is this needed?

brickbots (commented):
I'm very excited to have this nearby list!  Thanks for getting this rolling!

python/PiFinder/ui/object_history.py
mrosseel: indeed not, needed to be cleaned up

python/PiFinder/composite_object.py
mrosseel: can be used inline:
    def _filter_floats(self):
        return [float(x) for x in self.mags if is_number(x)]
        
agreed with the None

PR #208, opened 2024-08-10
Fix endless logging & collect early log messages in the log file

jscheidtmann:
There was a bug in Multiproclogging that could lead to an endless loop: 
 - logger.handle() was called for elements in the queue
 - which lead to an additional entry in the queue.

This is now fixed, by removing all handlers in the logging process and then creating a filehandler for writing to the logfile.


jscheidtmann:
You will probably see me requesting more changes to the log system after I return from vacation. :-D

PR #210, opened 2024-08-11
Fix logging crashes and recent bug

mrosseel:
- this fixes a few logging related crashes when running the software on a real pifinder
- this adds equality to the CompositeObject to fix a bug in Recent, which was introduced with making Recent a List instead of a custom object.


mrosseel:
more diffs than expected

brickbots:
@mrosseel Very nice to think about the __hash__ method as well as the __eq__ 🙌   Thank you for adding logging into the Pi Hardware specific modules

PR #213, opened 2024-08-12
Goto fix

mrosseel:
make goto work again


brickbots (approved):
Thank you for fixing the missing camera exposure while you were in here 💯

PR #221, opened 2024-08-25
Web Interface Location/Time entry screen asks for date Closes #220

### LINKED ISSUE #220: Web Interface Location/Time entry screen needs to ask for date
As discovered while testing #188 the web interface only requests the current time, assuming the system date is correct.  This will almost inevitably lead to incorrect results as the system date is very likely wrong

* Adjust web interface to include input for date along with time
* Change javascript to populate date as well from the browser
* Change back-end code (server.py line 211) to use provided date rather than system date

Vinnie-7:


brickbots:
Thank you @Vinnie-7!  This is much appreciated!  I'll merge it in and it will be part of our next release.

PR #222, opened 2024-08-26
Navigation bug between object details

mrosseel:


python/PiFinder/ui/object_details.py
```diff
@@ -387,7 +388,12 @@ def cycle_display_mode(self):
         self.update()
 
     def scroll_object(self, direction: int) -> None:
-        current_index = self.object_list.index(self.object)
+        if isinstance(self.object_list, np.ndarray):
```
brickbots, line 391: I'm going to merge this as it's fixing the issue, but it might be nice to figure out where this is coming from.  I'm guess it's a side effect of the nearby sort?

mrosseel: yes, nearby sort uses numpy arrays for speed. Should be revisited to either do all sorting in numpy or convert the nearby array

PR #228, opened 2024-09-06
Fixing magnitude filtering bug

mrosseel:
Three fixes which prevented succesful magnitude filtering. Also some extra nearby filtering calls (not sure if needed)


brickbots (approved):
Thanks for the additional comments here, for fixing the magnitude filter logic (pretty sure that bug was me) and for the general cleanup!

PR #229, opened 2024-09-06
Improve the behaviour of the Recents view

mrosseel:
If we are in the object details view for 10 seconds, the object is added to the recents view.
Exceptions: if you exit the objects view via something else than left/up/down then it won't register it. For instance long_press to main menu or long press to recent object.

Should this be added?


python/PiFinder/ui/menu_manager.py
```diff
@@ -358,6 +358,7 @@ def key_left(self):
         if self.marking_menu_stack != []:
             self.mm_select(self.marking_menu_stack[-1].left)
         else:
+            self.stack[-1].key_left()  # allow left intercept for recents
```
brickbots, line 361: It's funny, I've added this SAME functionality to support the new alignment system.... but one of the things I needed was to NOT have the UIModule unloaded after completion of the key_left work.  

I handled this by having the UIModule register its intent to take over the left button and the menu_manager checks this before unloading it if it's true calls the key_left() of the module.  

I think a unified way to do this would be to always call key_left of the module at the top of the stack, but have key_left return true/false to indicate if the menu_manager should unload the module.

brickbots: When I merge my new changes, I'll adjust :+1:

brickbots (approved):
Looks like a good first attempt here.  Have to see how it all works in practice, but it's better than just ignoring anything that gets scrolled through!

PR #231, opened 2024-09-07
Update testing scripts tooling

mrosseel:
This is not the menu-driven testing, just an update of the old monkey script machinery so that it works again.


jscheidtmann:
Would it make sense to invoke the 1k script from pytest?
Shouldn’t we add that as a smoke test?
(I have no clue how long the run time is.)

mrosseel:
> Would it make sense to invoke the 1k script from pytest? Shouldn’t we add that as a smoke test? (I have no clue how long the run time is.)

it takes quite a long time to run, also not sure how to test success.
In any case I still have issues with the test randomly hitting shutdown so need to find a fix for that still.

brickbots:
I was thinking about this last night and I think I have a solid plan for adding UIModule level testing into the PyTest framework.  This PR is more about running interactive automated keyboard bash testing, which is useful for the most high-level integration tests.

It does take a while though, and it requires a screen so can't be done right now on GitHub.  One of the first steps towards more complete testing will be to figure out how to use the Luma display stub class so that we can run tests without either the actual hardware or pygame.  I'll get an issue going with my ideas.

brickbots:
Here's my thoughts for more comprehensive UI testing.  I might start chipping away at this, but any assistance/volunteering is appreciated!

https://github.com/brickbots/PiFinder/issues/232

PR #236, opened 2024-09-12
Adjust Catalog Selection Logic, Add Filter by Constellation, Fix minor catalog defects

brickbots:
* Unify catalog selection attributes with filter. schema
* Make 'Select All/None' select all or no catalog across categories/sub-menus
* Add filtering by constellation - Present and working, but not exposed in menus.  
* Correct position of M45 in catalog
* Normalize object type data in catalog
* Add missing object types to filter selection and obj_type mapping


default_config.json
```diff
@@ -30,6 +30,113 @@
...
+        "Ast",
+        "Kt",
+        "***",
+        "D*",
+        "*",
+        "?",
+        "Pla",
+        "CM"
+    ],
+    "filter.constellations": [
```
mrosseel, line 50: I have a PR which has a more generic way of filtering, ie you can filter on any attribute of an object - when I get to it again I'll have to refactor,  but just a heads up that we don't have to hard code everything

brickbots: I'm curious what your filter refactor will look like here.  Does it operate on the principle of an empty list = all?

It used to be that an empty list (or None) of object types meant 'all', but I'm intentially switching the filtering system away from this as it required multiple exceptions and made the typing a bit strange.  Having the explicit list in the default_config is a nice way to know for sure what resetting the filters does.

PR #239, opened 2024-09-16
Fix wifi password with equals sign crash

### LINKED ISSUE #238: Populate_wifi_networks() in sys_utils fails when networks or password contain equal signs
Sys_utils.py line 50 uses string.split(„=„) which is wrong and fails, if any line contains additional equal signs.

mrosseel:
fixes #238

### DIFF
--- a/python/PiFinder/sys_utils.py
+++ b/python/PiFinder/sys_utils.py
@@ -1,6 +1,11 @@
 import glob
-import sh
-from sh import wpa_cli, unzip, su, passwd
+import re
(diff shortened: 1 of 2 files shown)


jscheidtmann:
+1 

I like the unit test!

brickbots:
Thanks for this quick fix of my bone-headed parser, and for breaking apart the file-read and parsing for effective testing 🙌 

Let's remove the two print statements (maybe log the error?) and get this merged :+1:

PR #248, opened 2024-10-21
[WIP] Adding WDS import, and a new way of handling object images

mrosseel:


brickbots:
I'm looking forward to adding more catalogs/objects.  We'll need to figure out something different with catalog images... maybe mark certain catalogs as not having/needing them?

PR #252, opened 2024-11-05
Add User Observing Equipment Config

brickbots:
Added Telescope and Eyepiece dataclass 
Added Equipment class containing a list of Telescopes and Eyepieces + active_telescope and active_eyepiece attributes.

Modified Config object to load/save/instantiate Equipment class.  Having the Config class hold this should make it avialble pretty universally and ties into the existing config file system.

Added some default equipment to the default_config.json file (200mm f5 dob + selection of Plossl eyepieces).

Todo:
- [x] Add some functions to compute eyepiece/telescope params, like True Field of View, Magnification, etc.

### DIFF
--- a/default_config.json
+++ b/default_config.json
@@ -49,43 +49,26 @@
             {
                 "make": "Generic",
                 "name": "Plossl",
-                "focal_length": 25,
+                "focal_length_mm": 25,
                 "afov": 50,
                 "field_stop": 21.2
             },
             {
                 "make": "Generic",
                 "name": "Plossl",
-                "focal_length": 11,
+                "focal_length_mm": 11,
                 "afov": 50,
                 "field_stop": 9.1
             },
             {
                 "make": "Generic",
                 "name": "Plossl",
-                "focal_length": 8,
+                "focal_length_mm": 8,
                 "afov": 50,
                 "field_stop": 6.5
             }
         ],
-        "active_telescope": {
-            "make": "Generic",
-            "name": "Dobsonian",
-            "aperture_mm": 200,
-            "focal_length_mm": 1000,
-            "obstruction_perc": 17.0,
-            "mount_type": "alt/az",
-            "flip_image": false,
-            "flop_image": true,
-            "reverse_arrow_a": false,
-            "reverse_arrow_b": false
-        },
-        "active_eyepiece": {
-            "make": "Generic",
-            "name": "Plossl",
-            "focal_length": 25,
-            "afov": 50,
-            "field_stop": 21.2
-        }
+        "active_telescope_index": 0,
+        "active_eyepiece_index": 0
     }
 }
--- a/default_config.json
+++ b/default_config.json
@@ -29,5 +29,63 @@
         "Str",
         "PL",
         "RDS"
-    ]
+    ],
+    "equipment": {
+        "telescopes": [
+            {
+                "make": "Generic",
+                "name": "Dobsonian",
+                "aperture_mm": 200,
+                "focal_length_mm": 1000,
+                "obstruction_perc": 17.0,
+                "mount_type": "alt/az",
+                "flip_image": false,
+                "flop_image": true,
+                "reverse_arrow_a": false,
+                "reverse_arrow_b": false
+            }
+        ],
+        "eyepieces": [
+            {
+                "make": "Generic",
+                "name": "Plossl",
+                "focal_length": 25,
+                "afov": 50,
+                "field_stop": 21.2
+            },
+            {
+                "make": "Generic",
+                "name": "Plossl",
+                "focal_length": 11,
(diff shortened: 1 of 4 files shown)


brickbots:
@WimDeMeester Here's the PR for the equipment tracking for your review.  Let me know if you have any questions or can see any areas for improvement!

WimDeMeester:
Thanks @brickbots,
I already tested a bit and everything seems to work fine.  I will add methods to compute magnification, true field of view, ...
I guess I can still change the eyepiece and telescope definitions (for example, I don't think that the field stop is really important information)?

WimDeMeester:
Will there be a method to update or remove an instrument / eyepiece?

brickbots:
They are lists, so the standard list methods will work.  The full round trip to make sure everything is saved:

```python
from PiFinder import config, equipment

config_object = Config()

new_eyepiece = equipment.Eyepiece(make="Generic", name="Plossl", focal_length=8, afov=50, field_stop=6.5)
config_object.equipment.eyepieces.append(new_eyepiece)
config_object.save_equipment()
```

Similarly,  you can use the various list methods like pop to remove by index or remove to remove by value

WimDeMeester:
Thanks!  Already used the example code for adding eyepieces, but I did not think about using the remove or pop to delete / update.

brickbots:
@WimDeMeester I've added mag and TFOV calculations as examples.  I also found an issue with serializing the active equipment and made some changes to make this more robust.  

I'm ready to merge this once you two are happy with it

WimDeMeester:
Looks good. 
I still have two questions: 
- Some people are using the PiFInder on a binocular.  Most of the time, these binoculars have a fixed magnification.  Can we add a fixed_magnification field to the telescope class?  This should also be taken into account in the calc_magnification method.
- Do we foresee to add a field maximum_magnification to the eyepieces class to enable zoom eyepieces?

brickbots:
Good questions and I'm not sure how best to handle these two edge-y cases.  

* I'm not really sure how many people would use a fixed eyepiece binocular with a PiFinder, but I'm sure someone will at some point.  For this, probably rare, case I'd propose that the user still create an eyepiece that represents the one which is permenantly installed in the binocular.  This avoids adding complexity for a very small, and potentially zero, set of users.

* Zoom eyepieces may be more popular, but I'm honestly not very familiar with them.  Even if we added a 'maxium magnification' we would need to also add some way for users to select which magnification they had the eyepiece set at to make some of the calculations work.  In this case, could a zoom eyepiece be represented effectively by a series of eyepieces?  Maybe the approach is to make a nice tool to add 3 (5?) eyepieces which cover the effective range of the zoom eyepiece?  We could add a sort of 'grouping' field to the eyepieces so they could be removed from the DB all at once?

brickbots:
I'm going to merge this for now as it's safe for any upcoming release to just have it there.  We can iterate on the idea freely in the context of the contrast reserve and I can start building it into some other areas like the image presentation / arrow switching

WimDeMeester:
> * I'm not really sure how many people would use a fixed eyepiece binocular with a PiFinder, but I'm sure someone will at some point.  For this, probably rare, case I'd propose that the user still create an eyepiece that represents the one which is permenantly installed in the binocular.  This avoids adding complexity for a very small, and potentially zero, set of users.

I'm anyway using the PiFinder with my 20x80 binoculars.  I need to think how I can add the correct eyepiece (as I don't know the focal length of the binoculars at all).  But I guess I can find the information somewhere on the internet. 

> * Zoom eyepieces may be more popular, but I'm honestly not very familiar with them.  Even if we added a 'maxium magnification' we would need to also add some way for users to select which magnification they had the eyepiece set at to make some of the calculations work.  In this case, could a zoom eyepiece be represented effectively by a series of eyepieces?  Maybe the approach is to make a nice tool to add 3 (5?) eyepieces which cover the effective range of the zoom eyepiece?  We could add a sort of 'grouping' field to the eyepieces so they could be removed from the DB all at once?

Zoom eyepieces are indeed much more complex to add, and I think they should not be added yet.  

Thanks for the merge!  This will certainly do for now.  I'm in the process of adding a page to the web interface to add new instruments and eyepieces.  As soon as I'm ready, you can expect a Pull Request for this.  
Next steps are then:
- A selection screen on the PiFinder to change telescope / eyepiece.
- Show the contrast reserve for a given object, using the given telescope / eyepiece combination.  Of course, to be able to do this, we also need to have the SQM value of the sky (fixed in the settings of the PiFinder or even better, calculated from the pictures taken by the PiFinder).
- Next step will be to add the contrast reserve to the list, so that the user can sort on contrast reserve.

brickbots:
Hi @WimDeMeester Yes, good catch!  This is merged into main so I made a small commit there to switch this to float :+1:

PR #254, opened 2024-11-07
build guide small improvement

mrosseel:


brickbots:
Read the Docs allows you to keep builds of older versions, so I think we are covered for the user_guide_old.  You can see an example of this here:
Screenshot 2024-11-09 at 1 32 41 PM

If you want to close this PR, I'm going to be merging the v2.5 camera upgrade docs with the ones here and I can make the left/right change as well :+1:

mrosseel:
> Read the Docs allows you to keep builds of older versions, so I think we are covered for the user_guide_old. You can see an example of this here: Screenshot 2024-11-09 at 1 32 41 PM
> 
> If you want to close this PR, I'm going to be merging the v2.5 camera upgrade docs with the ones here and I can make the left/right change as well 👍

Ok

PR #256, opened 2024-12-03
Update Cedar Detect / Solve and Solidify Align

brickbots:
* Added new migration system to avoid re-running post software update code
* Switched tetra3 repo to cedar version at release 0.5.1
* Switch to updated binaries for cedar detect
* Align function now uses affordances from solver/detect for speed and reliablity
* Make new align the default

After merging to main this will require running pifinder_post_update script in any development repos to accomplish:
* tetra3 repo switch
* Updated requirements

PR #259, opened 2024-12-20
Add Architecture documentation

jscheidtmann:
I have created some architecture documentation that should help new developers find their way around. 

This is open:
- [x] Include feedback from @brickbots 
- [x] Include feedback from @mrosseel 
- [x] Create a nice drawing of the collaboration of processes (Jens)

(Please ignore the branch name, for the time being.)


docs/source/dev_arch.rst
```diff
@@ -68,18 +68,75 @@ This choice of architecture means that logging to disk is a little bit more comp
...
 ------------------------ 
 
-PiFinder uses 
+PiFinder uses ``cedar-detect-server <https://github.com/smroid/cedar-detect>``_ 
+in binary form to determine star centroids in an image. This is a fast centroider written
+in the Rust programming language that is running in a separate process. A gRPC API is used
+to interface with this process. 
 
+The detected centroids are then passed to the 
+``tetra3 solver <https://github.com/esa/tetra3>``_ for plate-solving.
```
smroid: This is not quite accurate. The plate solving is done by Cedar-solve, which is a fork of Tetra3.

brickbots: We should definitely update this text as that will be the case by the time this is released. This switch to Cedar Solve will be done with PR #256 which I'll be merging in over the break and will be part of the same release this PR will be a part of.  :+1:

smroid: Ah, I didn't realize PiFinder is still using o.g. Tetra3.

brickbots: Yeah... I've been... slow on the switch 😅  Mainly because I need to make sure the whole sub-repo switch, new binary for detect, etc all work during the software update without borking everything.  The PR mentioned seems to do this properly so the next software update will be where we *finally* make the full switch 🎉

brickbots: Oh, and to clarify, it's using a commit you PR'd to the tetra3 repo, so it is using effectively an early version of Cedar Solve 🙌  Just before it was a fork

jscheidtmann: Ok, I'll check the respective PR.

jscheidtmann:
Hi, I added some more documentation, including a generated picture of how processes are associated. 
I also took PR #256 into account.

brickbots:
Thank you!!!  I'm loving the mermaid chart in there.  Would you please check in the source for that as well? 

Relatedly, did you know that Github will render Mermaid charts! 

In this case we still need the PNG for readthedocs, but I was really happy when I realized github will render charts embedded in markdown 😄

jscheidtmann:
It’s already in: The mermaid source is hidden in the restructuredText as a comment, right above the image. I can put that in a separate MD file tomorrow, if you want.

brickbots:
No, that's perfect, I just missed it there!  Thank you ✨

Issue #129, opened 2024-01-25
Investigate contrast reserve


mrosseel:
In deepskylog.be, they implemented a Contrast Reserve.
This takes into account your local SQM, telescope and magnification to calculate how easily visible an object would be.

Here is the PHP implementation:
https://github.com/DeepskyLog/laravel-astronomy-library/blob/edf28c1ff103dc0822a72b7902ba0c00fba8ba89/src/deepskylog/AstronomyLibrary/Targets/Target.php#L1287

brickbots:
This is a really cool concept.  I very much like the idea of a dynamic visibility measure and I've been dreaming of calculating SQM based on the camera capture.  I'm pretty sure this is possible (there are cell phone apps which do this) and should be reliable since the imaging system is very uniform (unlike cell phones).

WimDeMeester:
Hi @mrosseel , @brickbots 
I will create a python library to calculate the contrast reserve of a given object (how easy / difficult / impossible it is to see), given the SQM, the telescope diameter and the magnification (or eyepieces).  
Of course, some changes will be needed in the piFinder software too.  I would be good if:
- The SQM is calculated from the solved images (see #133).
- The telescope (focal length) can be added in the piFinder.
- A list of eyepieces can be added in the piFinder.

I would be wonderful if different configurations (telescope / eyepiece combinations) could be saved (and used) in the PiFinder software.
I will probably only have time to work on the python library in September...

mrosseel:
hi @WimDeMeester great to hear you're planning to work on the contrast reserve! 
As you say there are some missing pieces that need to come together to make this work. Telescope and eyepiece combos have been on the wish list for a while (to scale the images we show per object for instance), contrast reserve might be the push that was needed.
Regarding timeframe, there are some features we want to finish over the summer so September sounds fine :)

WimDeMeester:
Ok, I did find some time this week.
I have created a python package to calculate the contrast reserve for an object.  The package can be found here:

https://github.com/DeepskyLog/pydeepskylog

Installation is possible using 

`pip install pydeepskylog`

Updating PiFinder to use this library will be for September (earliest).  I will need to think a bit on how to integrate this in the best way into PiFinder and I will probably need some help too :P

brickbots:
I'm very excited about this concept and implementation @WimDeMeester !  One of the things that has been on the PiFinder todo list for a while is to allow people to configure parameters of their scope and eyepieces.  We were planning on using this information for:
[...]

PR #261, opened 2024-12-23
[WIP] Contrast reserve

WimDeMeester:
Adds the calculation of contrast reserve to PiFinder.
The value of the contrast reserve is shown in the object detail page:
2024-12-23 15_12_20-Luma core Display Emulator (piFinder)

Remarks:
- [ ] As the SQM is not yet calculated by the PiFinder, I take a fixed value of 20.15 for the SQM.
- [ ] The contrast reserve is just a number... For most users, this number does not mean anything.  We still need to find out how to display the contrast reserve in a more meaningfull way

| Contrast Reserve | Visibility             |
|------------------|------------------------|
| < -0.2           | Not visible            |
| -0.2 < CR < 0.1  | Questionable           |
| 0.1 < CR < 0.35  | Difficult              |
| 0.35 < CR < 0.5  | Quite difficult to see |
| 0.5 < CR < 1.0   | Easy to see            |
| 1.0 < CR         | Very easy to see       |

This pull request will solve #129

### DIFF
--- a/python/PiFinder/ui/object_details.py
+++ b/python/PiFinder/ui/object_details.py
@@ -5,7 +5,7 @@
 This module contains all the UI Module classes
 
 """
-
+import pydeepskylog as pds
 from PiFinder import cat_images
 from PiFinder.ui.marking_menus import MarkingMenuOption, MarkingMenu
 from PiFinder.obj_types import OBJ_TYPES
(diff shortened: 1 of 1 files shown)


jscheidtmann:
I take it it's the number in the upper right?

Some ideas on the "just a number":
- Use a bar graph: `|===|=* |   |   |`
- Use a categorical scale: `--, -, 0, +, ++`
- Zero to 5 stars: `|***  |`  
- ...

WimDeMeester:
> I take it it's the number in the upper right?
> 
> Some ideas on the "just a number":
> 
> * Use a bar graph: `|===|=* |   |   |`
> * Use a categorical scale: `--, -, 0, +, ++`
> * Zero to 5 stars: `|***  |`
> * ...

Good ideas!  I will think a bit more about it during the christmas break.

mrosseel:
Hi Wim, you could just convert the CR to a piece of text that is added to the description. Where is probably the biggest question, i.e in front of the object description or at the end. Something like 'Visibility: easy'

PR #263, opened 2025-01-02
Extend Equipment Functionalty to Object Image Preview

brickbots:
This PR switches the object image preview FOV to calculated values based on current eyepiece/telescope combinations and allows cycling through eyepieces with the +/- key while viewing object details.

To enable this and help reduce code duplication I've also changed the Equipment and Logging eyepiece selection menu to use a common Eyepieces menu via the jump functionality.  This menu is generated at start up based on the eyepieces available at start-up time.

Changes:
* Fixed up bug in logging observations due to change to Cedar by adding a utility function to serialize solution which handles numpy objects
* Added jump function callback as parameter to UI/menu object init
* Added generation of eyepiece menu during menu manager init
* Pointed logging and equipement eyepiece menu to generated menu_structure eyepiece menu via jump functionality
* Adjusted cat image utility to accept general eyepiece name for burn in rather than image souce
* Made cat image always look for POSS, in the future it would be good to have one 'best' image for each object
* Cat imgage fov is based on selected eyepiec
* Added eyepiece selection cycle to equipment object

TODO:

- [x] Trigger regeneration of eyepiece menu when new eyepieces are added via the web interface


brickbots:
@WimDeMeester Pinging you in case you want to check some of the updates to the equipment class and UI module 👍

WimDeMeester:
> @WimDeMeester Pinging you in case you want to check some of the updates to the equipment class and UI module 👍

@brickbots I can see that the preview image of the object is now using the selected eyepiece.  By pushing the + or - buttons when I selected the image preview, I can change the used eyepiece, very nice!  Is the preview also using the flip and flop options?  

Two remarks:
- I think I can not change the telescope anymore from the Tools/Equipment menu.  Can you have a look? 
- I guess we should also add the telescope to the observations.  I think the eyepiece is added correctly.

brickbots:
@windemeester I've fixed up the linkage between the web interface and the onboard UI.  Now changing eyepieces/telescopes via the web interface is immediately visible in the onboard UI.

Adding/Deleting equipment does require a restart to update the menus, but I think this is an okay requirement for now :+1:

Flip/Flop and Arrow config per instrument are not implemented yet.  I've created issue #270 and assigned it to myself for the next release

PR #271, opened 2025-02-03
Update DeepskyLog import of instruments

WimDeMeester:
I have updated DeepskyLog to add the missing fields like

- obstruction
- Flip / flop
- Mount
- Make

This Pull Requests makes sure that these fields are also used when importing from DeepskyLog.


brickbots:
Looks great!  I'm excited to dig more into DeepskyLog and I appreciate you fixing up my typo in the menu_structure

PR #280, opened 2025-02-14
Fix empty console bug.

jscheidtmann:
Make UIConsole a single instance class, so that the console that is created in main.py is picked up by the menu_manager, when the Tools->Console menu item is invoked (which usually creates a new instance).


jscheidtmann:
@brickbots please review!

PR #281, opened 2025-02-15
Replay of ubx files from fake GPS. 

jscheidtmann:
- Using `-fh` gps_fake.py now reads from test_ubx/*.ubx files and replays these to PiFinder (one message per second). 
- Added a logconf_gps.json log configuration file that enables maximum output from the GPS subsystem (change symbolic link pifinder_logconf.json to use it).
- Cleanup of logging and log-levels in GPS subsystem (to use logger only and only logger).


brickbots:
How long does this now take to get a lock using the reply?  Most of the time I use -fh to test some other feature and having to wait for a replay GPS lock might not be helpful.

Perhaps we need to set up a specific command line argument to determine which GPS module to load?  Perhaps do away with -fh all together in favor of specific hardware emulation options?

--camera
--imu
--gps
--keyboard
--screen

It would be a lot of typing, but we could have some sensible defaults or shell scripts?

jscheidtmann:
At the moment all files in test_ubx/ are replayed. If you stored `a.ubx` while there was a lock, and it is the only file, this is going to replayed all the time. So the lock would be immediate.  

How about adding options to -fh? Such as `gps-lock`, `gps-replay`,  `real-imu`, and so on.

jscheidtmann:
@mrosseel I'll add the M10 changes here, ok?

mrosseel:
Can you just branch from this and make a new PR?

jscheidtmann:
Will do

mrosseel:
Usually I number the prs so it's clear the order in which they should be
merged. I need to test your pr still but once the ubx changes are in I can
look at it

jscheidtmann:
I wanted to implement the suggestion with the command line parameter and I probably have to create a new PR to make it go against main now, haven't I?

jscheidtmann:
This PR is continued in #305

Issue #286, opened 2025-02-19
Logging seems to be split


jscheidtmann:
DeadricGems reported a hang after a failed upgrade in Discord at this point:

![Image](https://github.com/user-attachments/assets/011ea068-d71a-4ecd-951a-9b13666dc248)

But the log file, that we could access via SMB was empty.

I seems that the log configuration is switched / wrong at that point in the boot process.

jscheidtmann:
I cannot find that "WiFi:Client" message in the source. Not even find `find . -type f -print0 |xargs -0 grep -i wifi:client`.

brickbots:
The splash screen is handled in the file splash.py which is called via a different start up service before networking and other bits are online.

Even if it crashes, the main code should start up, so I don't think having proper logging on the splash screen is needed....

This still does not really help with why the software ended up in an un-usable state during the update and the main.py file crashed before logging anything 🤔   Given that the state of the code/sd card could end up very corrupted if the updated is terminated part way through I'm not surprised it can end up in some broken conditions.

Making the update process more robust, with a fail-safe rollback, is definitely needed

PR #287, opened 2025-02-19
Better logging in an attempt to catch #286

jscheidtmann:
- Checked logging and all messages from the start are logged. 
- Added log statements for starting of sub processes. 
- Be more chatty for GPS and Solver.
- Catch value error, that gets thrown by newer versions of cedar (instead of a FileNotFound)
- Display messages in console


brickbots:
This is a nice improvement, and I'm happy to merge it!  Your point about essentially wanting to log each time a console message is emitted is a good one.  They should probably just be the same: #290

PR #289, opened 2025-02-19
Fix crash when adding instrument with a float as aperture or focal_length

WimDeMeester:
Fixes the crash that appears when adding an new instrument with aperture (or focal length) as a float instead of an integer.

@brickbots Can you add this to the next release?

### DIFF
--- a/python/PiFinder/server.py
+++ b/python/PiFinder/server.py
@@ -502,8 +502,8 @@ class Server:
                 instrument = Telescope(
                     make=request.forms.get("make"),
                     name=request.forms.get("name"),
-                    aperture_mm=request.forms.get("aperture"),
-                    focal_length_mm=request.forms.get("focal_length_mm"),
+                    aperture_mm=int(request.forms.get("aperture")),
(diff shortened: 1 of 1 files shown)


jscheidtmann:
I can still enter invalid numbers in Chrome, like "123.345e4443". The message displayed then says "Instrument added, restart PiFinder to use". 

the error message in the log is: 

2025-02-20 07:46:55,497 Webserver-Server:ERROR:Error adding instrument: invalid literal for int() with base 10: ''

jscheidtmann:
Further note that on Safari (iPhone iOS 18.3.1) you can still enter other characters. 
It seems that it's not crashing anymore.

WimDeMeester:
In the html input element, I have put number and 123.345e4443 is indeed a number, meaning 123.345 times 10 to the power 4443.  Converting this to an integer will indeed not work, because an integer can not store such a large number.  Maybe I can limit the aperture to some reasonable value, but the question is indeed what a reasonable value is...

brickbots:
Thank you for helping to make this more robust!

PR #292, opened 2025-03-09
Unify gps status checking

brickbots:
This PR brings in the GPS data structure refactor from pr#268 and:
* Add altaz_ready boolean to shared state - This is the preferred way to check that both time and date are available and accurate.
* Use altaz_ready in various places that used to check location/datetime in various combinations to avoid differing results
* Adjust planet catalog init/scheduling - as initial location is no longer set from config, planet catalog gen needs to wait.  altaz_ready is checked every 10 seconds to trigger initial planet catalog gen, then 307 second thereafter to update planet positions
* switch all usage of initialise to initialize - it's an arbitrary regional difference, but we have to choose one :-) 
* timezone calculation is moved to shared state to make sure it's triggered whenever location is set
* test mode now uses messages trigger centralized location/time setting in the main module

@mrosseel This may have had the side effect of preventing the solver crash when setting location/datetime from the web interface.  I've tested this a lot and it seems to not crash any longer 🤷

### DIFF
--- a/default_config.json
+++ b/default_config.json
@@ -1,13 +1,6 @@
 {
     "display_brightness": 255,
     "keypad_brightness": "0",
-    "last_location": {
-        "lat": 34.0,
-        "lon": -118.0,
-        "altitude": 51.0,
-        "gps_lock": false,
-        "timezone": "America/Los_Angeles"
-    },
     "camera_exp": 400000,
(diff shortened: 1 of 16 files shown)


mrosseel (approved):
didn't test, but changes seem correct

PR #305, opened 2025-03-31
Gps ubx replay with fake hardware

jscheidtmann:
If -fh is enabled the file ./test_ubx/test.ubx is replayed as fast as possible on repeat. 
If that file does not exsists, a fix and it's homing in is simulated.

You can call `python -m PiFinder.gps_ubx_parser <file>` to parse and print out ubx messages on the command line. 

This PR fixes the problem, that modern GPS chips work with 10Hz or faster and messages pile up on the gps_queue.
This PR also fixes the problem that on modern chips the "sats used" display in Status is 0 (zero) all the time.

This is in continuation of PR #281, but this time against main, into which gps-ui was merged. 

Due to the new GPS locking logic, it does not make sense to cycle through files. Instead only test.ubx is replayed. Make that a symlink to file with the behavior you want. 

We planned to do this, but it does not make sense anymore: 

- Using -fh gps_fake.py now reads from test_ubx/*.ubx files and replays these to PiFinder (one message per second).
- Added a logconf_gps.json log configuration file that enables maximum output from the GPS subsystem (change symbolic link pifinder_logconf.json to use it).
- Cleanup of logging and log-levels in GPS subsystem (to use logger only and only logger).

At the moment all files in test_ubx/ are replayed. If you store a `a.ubx` while there is a lock, and it is the only file, this is going to replayed all the time. So the lock is immediate in that case.

- [ ] Parameter „lock“, simulate immediate lock
- [ ] Parameter „sim“, simulate the gps getting on target
- [ ] Parameter &lt;file glob&gt;, that replays matched files from test_ubx/
- [ ] `-fh` not taking care of GPS.


jscheidtmann:
Please give it a try, before merge with actual units

d33psky:
image
works great !

brickbots:
Thanks for the testing @d33psky and for the helpful debug feature @jscheidtmann 🤝

PR #306, opened 2025-03-31
Web logging

mrosseel:
- add a logs entry to the web interface to see logs, change log levels and download logs. For better debugging.


python/PiFinder/gps_ubx_parser.py
```diff
@@ -141,19 +141,23 @@ async def connect(cls, log_queue, host="127.0.0.1", port=2947, max_attempts=5):
         attempt = 0
         while attempt < max_attempts:
             try:
+                # Add delay between connection attempts
+                if attempt > 0:
```
jscheidtmann, line 145: Seems as if this commit belongs into another PR…

jscheidtmann:
A log configuration can be different levels for different loggers. I believe because of this, a „current_level“ may not be meaningful. It would make more sense, to resolve the symbolic link `pifinder_logconf.json` and give the name of the configuration file for logging. If we want to provide more infos, we could display the first 10 lines or so of that config file (which could contain a description). 

Re /log/components:
There‘s a lib which outputs a tree: https://pypi.org/project/logging-tree/
This could maybe on a separate page? 

Note: I have not tried running this, yet.

PR #311, opened 2025-04-06
[WIP] More secure WiFi configuration

jscheidtmann:


brickbots:
Hi @jscheidtmann Thank you for working on this!  I think the overall plan is well thought out, but the final TODO about suggesting ethernet is not going to work as most PiFinders don't have the ethernet port exposed and many people don't have the cabling anyway.  

It's very true that ethernet would ultimately be more secure, but I don't think setting up wifi over a non-encrypted HTTP connection is much of a concern as it's not traversing the internet.  To be intercepted, someone would need to be local to the person and sniffing traffic on the wifi network generated by the PiFinder.  While such an attack is possible, being in the right place at the right time to pull this off would be exceedingly difficult and the payoff would be pretty minor.

brickbots:
A really interesting option here would be to display a QR code with the PiFinder WiFi SSID/Password on the screen. Correctly typing a 20 char random password might be frustrating on a phone or tablet during initial set up.

Here's a library that seems like it would do the job, or could be modified if needed:
https://pypi.org/project/wifi-qrcode-generator/

mrosseel:
hi @jscheidtmann I'm having doubts on the general approach here. 
- I like the random hostname
- the random 20 character password could be a support nightmare, if the UI fails to start there's no way to ssh into the pifinder and troubleshoot. Reading a 20 char password from the small screen will also go more wrong than right. QR code could be a solution here, but we still have the no ui trouble.
- when accessing via the web interface you need the password (each time) but you can't navigate to that screen without the password

maybe @brickbots has some input here as well

jscheidtmann:
First note: The code only triggers, when "CHANGEME" is present in the SSID. We could make it triggering on "PiFinder-default-CHANGEME" only or some such. 

Brickbots wrote: 
> Hi @jscheidtmann Thank you for working on this! I think the overall plan is well thought out, but the final TODO about suggesting ethernet is not going to work as most PiFinders don't have the ethernet port exposed and many people don't have the cabling anyway.

I know, I was suggesting to use ethernet _during the build process_ to define e.g. client WiFi entries and avoid entering home wifi password over an twice unencrypted link (open wifi & http). 

> It's very true that ethernet would ultimately be more secure, but I don't think setting up wifi over a non-encrypted HTTP connection is much of a concern as it's not traversing the internet. To be intercepted, someone would need to be local to the person and sniffing traffic on the wifi network generated by the PiFinder. While such an attack is possible, being in the right place at the right time to pull this off would be exceedingly difficult and the payoff would be pretty minor.

The http and not httpS is not a problem, as long as the WiFi itself is encrypted, which by default it is not.  I am more concerned about entering the home WiFi password over an entirely unencrypted link. As we are addressing a pretty nerdy group of DIYs, that means they are usually also quite security concious, and this is especially so in Germany.  

I usually have no problem to use the Pi unencrypted, although people could make fun of you on a star party (I know some of those guys, who would). 

MRossel worte:
> hi @jscheidtmann I'm having doubts on the general approach here.
> 
> * I like the random hostname

Good. That's in especially for star parties. 

> * when accessing via the web interface you need the password (each time) but you can't navigate to that screen without the password

No, this long string is the WiFi password, which will be stored by your phone / tablet. The password that needs to be entered in the web is pifinder's (the unix user). This behavior is not changed.   

> * the random 20 character password could be a support nightmare, if the UI fails to start there's no way to ssh into the pifinder and troubleshoot. Reading a 20 char password from the small screen will also go more wrong than right. QR code could be a solution here, but we still have the no ui trouble.

- I believe using a 12 random char password would be enough (that's what Microsoft recommends as a minimum). 
- Yes, you have a point there when the UI is not coming up. This means having to ssh into the PiFinder. I do not have an immediate solution to this.

jscheidtmann:
This is an example password display: 
image

jscheidtmann:
Display of WiFI QR Code: 
image

jscheidtmann:
Regarding the UI not coming up: 
We could implement a watchdog, that looks at the status of the PiFinder service every 10 seconds or so. If the status is "failed" not started or some such, it runs a script, which then displays the WiFi QR code on the display.

python/PiFinder/ui/wifi_password.py
```diff
@@ -0,0 +1,297 @@
...
+
+            self.marking_menu = MarkingMenu(
+                left=MarkingMenuOption(
+                    label="QR", callback=self.mm_display_qr, enabled=True
+                ),
+                right=MarkingMenuOption(
+                    label="Passwd", callback=self.mm_display_pwd, enabled=True
+                ),
+                down=MarkingMenuOption(
+                    label="Exit", callback=self.mm_switch_mode, enabled=True
```
brickbots, line 56: There is a built in marking menu system to make this easier:

`MarkingMenuOption(label="SwitchMode", menu_jump="wifi_mode")`

jscheidtmann:
@brickbots @mrosseel The feature is complete now, if you could please have a look.

jscheidtmann:
image

jscheidtmann:
Tested the functionality of this on a real PiFinder:
-[x] Open network displays QR code and "<No password defined>". QR Code works.
-[x] encrypted network display QR code and the password, QR code works.
-[x] adding CHANGEME to AP name created a random AP name 
-[x] adding ENCRYPTME to AP name adds encryption on start of service.
-[x] you can add CHANGEME and ENCRYPTME at the same time
-[x] network dialog displays error messages if password is too short.
-[x] network dialog can be used to enable encryption.

jscheidtmann:
This PR closes #179

jscheidtmann:
- [x] Test password of AP can be changed.

jscheidtmann:
Note: It might be the case, that you can see a "PiFinder-CHANGEME(-ENCRYPTME)" in the WiFis of your phone or tablet.

jscheidtmann:
Functionality tested successfully. Please review

PR #324, opened 2025-06-25
Show magnification

mrosseel:
In the image view, not only show the FOV but also the magnification of the chose eyepiece.
Works both if there is an image or not (this was previously not the case).
Only open point is that the circle is now centered, and partially occluded by the title bar. Maybe we should shift it down so both top and bottom parts of the circle are cut off a little, or just resize the circle.

Screenshot 2025-06-25 at 16 26 30

Screenshot 2025-06-25 at 16 30 04


brickbots:
Thank you!!!  We might consider adding some option for hiding the title bar?  Maybe auto-hide in some cases?  In any case I don't mind the now centered circle!

jscheidtmann:
By the way, does the context menu (long press of square) have a change eyepiece entry?

PR #329, opened 2025-06-29
Fix Debug Dump and split "Save Picture" into separate hotkey

jscheidtmann:
- refactor keyboard_local to have a separate key for all key codes from keyboard_interface
- Add hotkey ALT_LEFT = SQUARE+LEFT to save current image and split that from debug dump.
- Fix debug dump to not throw exceptions.
- Tested using Test mode and saving image / debug dumping


jscheidtmann:
Hi @brickbots, this is ready to be merged. Can you have a look? Thx

PR #331, opened 2025-07-13
Migrate Webserver to Flask and Jinja templates.

jscheidtmann:
... from Bottle.

This PR migrates the internal PiFinder webserver to Flask and Jinja and translates the webserver strings.

- [x] Created selenium tests for all webserver functionality
- [x] Successfully ran selenium test suite against server2 webserver (jinja2 templates)
- [x] have babel extract the newly identified strings
- [x] Translate strings to German
- [x] Translate strings to French
- [x] Bootstrap spanish translation using LLM
- [x] Remove Indi PoC
- [x] A/B/C/D -> LEFT/UP/DOWN/RIGHT etc.
- [x] Remove server.py/view and rename server2.py/views2
- [x] Move driver and shared_driver to web_test_utils
- [x] Remove bottle dependency
- [x] Run the tests against an actual unit
- [x] Tune down webserver logging to not swamp the online logs display
- [x] Test with Safari browser on actual unit, if downloads work.

This PR introduces a /api/current-selection endpoint that provides information on the currently selected menu entry or ui-element. The selenium tests use this to validate the remote control and _change_ the state of the PiFinder.

Known limitations:
- The language of the webserver does not change when the language in the UI is changed. This is due to the webserver running in it's own process. At the moment, you need to restart the PiFinder service, so that the new language is picked up (while the UI on the PiFinder picks up the change immediately).
- The selenium tests need to have the PiFinder using english language.


brickbots:
Hi @jscheidtmann Thank you for this switch to enable i8n and the great test coverage 🙌  This is a great set of improvements.  I have a couple requests and a couple of questions I'll put here and I'll leave a couple of inline notes:

Change Requests:
* I think the Indi POC commit may have snuck into the wrong branch.  Would you please remove it in prep for merging this PR?
* In both server and server2 the key codes were switched back to UP/DN/A/B/C/D from PLUS/MINUS/UP/DOWN/LEFT/RIGHT.  I can see that this was fixing a bug caused by the template still sending UP/DN/A/B/C/D.  To correct this defect I'd rather keep the key codes using PLUS/MINUS/UP/DOWN/LEFT/RIGHT and adjust the templates to match.  

Questions:
* Why server2/views2 instead of replacing server/views?  Is there a reason to have both available?  If not, let's switch over and not duplicate the code/views.
* The UI State serialization adds a lot of code in the menu manager and UI modules.  I can see this is useful for testing... but I can't help thinking we can get almost the same testing capabilities with less code.  I guess I don't really have a question here... just a note for future tech/debt reduction.  I'm going to file an issue to revisit this once the main code is AsyncIO enabled.  Once this is done, we can run a server/endpoint specifically for testing from within the main UI code and potentially be a bit more specific about the items we serialize, or re-work the UIBase class to use introspection to do a bit of auto-serialization. 👍

python/tests/website/test_web_interface.py
```diff
@@ -0,0 +1,141 @@
+import pytest
+from selenium import webdriver
+from selenium.webdriver.common.by import By
+from selenium.webdriver.chrome.options import Options
+from selenium.webdriver.support.ui import WebDriverWait
+
+
+@pytest.fixture
+def driver():
```
brickbots, line 9: Can we migrate this driver which is used across many test modules into the web_test_utils.py module?

python/tests/website/test_web_equipment.py
```diff
@@ -0,0 +1,731 @@
...
+Data Persistence: Verifies that equipment data persists correctly across page refreshes and navigation.
+Clean Test State: Tests clean up after themselves by removing test entries to avoid interference between runs.
+
+Infrastructure: Uses the same Selenium Grid setup as other web tests with automatic skipping when unavailable.
+(Summary created by Claude Code)
+"""
+
+
+@pytest.fixture(scope="session")
+def shared_driver():
```
brickbots, line 59: Can this also be moved into the web_test_utils.py to be used across all the test modules?

brickbots:
Thanks again @jscheidtmann!  Please let me know if you have any questions or thing I'm off base on any of these comments 🤝

brickbots:
Thanks @jscheidtmann! 

Yes, I'm also not sure what to do with the UI testing code.  It's good to have, and it should not cause much issue at runtime, but it does add to the source file size and probably does impact performance on the main loop at least a little bit.  I think we leave it for now to get this all merged and I can circle back when I have some more time to think about it 👍 

I've created a ticket so I don't forget: #346

jscheidtmann:
moved todo list to first comment.

PR #335, opened 2025-07-30
Pi HQ camera light sensivity improvement

c3n:
I'm using a Pi HQ camera M12 with a cheap 25mm lens (approx. f/2). Even with 1s exposure some places I didn't get a lock, especially near the horizon  (the cedar-server image locked really quickly and without problems). I changed the analog gain to 22 (like in cedar server), added a line of code to subtract the offset (around 256) and am now using the lower 8 bits instead of the upper 8 bits of the raw image. Now it locks beautifully with 0.4s (0.2s works about as good as 0.8s used to).

### DIFF
--- a/python/PiFinder/camera_pi.py
+++ b/python/PiFinder/camera_pi.py
@@ -27,20 +27,36 @@ class CameraPI(CameraInterface):
         from picamera2 import Picamera2
 
         self.camera = Picamera2()
-        # Figure out camera type, hq or imx296 (global shutter)
-        self.camera_type = "hq"
-        self.gain = 22
+
         self.exposure_time = exposure_time
+        self.format = "SRGGB12"
         self.bit_depth = 12
+        self.digital_gain = 1.0 # TODO: find optimum value for imx296 and imx290
+        self.offset = 0 # TODO: measure offset for imx296 and imx290
+
+        # Figure out camera type, hq or imx296 (global shutter)
         if "imx296" in self.camera.camera.id:
             self.camera_type = "imx296"
+            # The auto selected 728x544 sensor mode returns black frames if the
+            # exposure is too high
+            self.raw_size = (1456, 1088)
+            self.format = "R10"
+            self.bit_depth = 10
             # maximum analog gain for this sensor
             self.gain = 15
-            self.bit_depth = 10
-
-        if "imx290" in self.camera.camera.id:
+        elif "imx290" in self.camera.camera.id:
             self.camera_type = "imx462"
+            self.raw_size = (1920, 1080)
             self.gain = 30
+        elif "imx477" in self.camera.camera.id:
+            self.camera_type = "hq"
+            # using this smaller scale auto-selects binning on the sensor
+            self.raw_size = (2028, 1520)
+            self.gain = 22 # cedar uses this value
+            self.digital_gain = 13.0 # initial tests show that higher values don't help much
+            self.offset = 256 # measured with lens cap on, matches what the internet says
+        else:
+            raise Exception(f"Unknown camera type: {self.camera.camera.id}")
 
         self.camType = f"PI {self.camera_type}"
         self.initialize()
@@ -48,29 +64,15 @@ class CameraPI(CameraInterface):
     def initialize(self) -> None:
         """Initializes the camera and set the needed control parameters"""
         self.stop_camera()
-        if self.camera_type == "imx296":
-            # The auto selected 728x544 sensor mode returns black frames if the
-            # exposure is too high.  So we need to force a specific sensor
-            # mode by specifying a raw stream we won't use
-            cam_config = self.camera.create_still_configuration(
-                {
-                    "size": (512, 512),
-                },
-                raw={"size": (1456, 1088), "format": "R10"},
-            )
-        elif self.camera_type == "imx462":
-            cam_config = self.camera.create_still_configuration(
-                {
-                    "size": (512, 512),
-                },
-                raw={"size": (1920, 1080), "format": "SRGGB12"},
-            )
-        else:
-            # using this smaller scale auto-selects binning on the sensor...
-            # cam_config = self.camera.create_still_configuration({"size": (512, 512)})
-            cam_config = self.camera.create_still_configuration(
-                {"size": (512, 512)}, raw={"size": (2028, 1520), "format": "SRGGB12"}
-            )
+        cam_config = self.camera.create_still_configuration(
+            {
+                "size": (512, 512)
+            },
+            raw={
+                "size": self.raw_size,
+                "format": self.format
+            },
+        )
         self.camera.configure(cam_config)
         self.camera.set_controls({"AeEnable": False})
         self.camera.set_controls({"AnalogueGain": self.gain})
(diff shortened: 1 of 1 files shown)


mrosseel:
interesting change!
Some questions:
- Do you have any insight in how general this offset is (across pihq sample variation and temperatures) ?
- using lower bits might mess with photometry (like for an SQM measuring feature) but it for sure helps sensitivity

Could you point this PR to the brickbots:main branch, which is were new stuff lands before being released? So it can be tested before releasing in the wild

brickbots:
Hi @c3n Thanks for the work on this!  Having the HQ camera perform at their best is definitely something we're interested in.  

In looking at your code, I think the main improvement in solving is the removal of the offset before converting to 8 bit, but I think you are losing some overall signal range when you just clip the values rather than scale into the 8 bit space.

For the imx477 code path:
raw_capture is already converted to int16 on line 109 as part of the crop.  The sensor is 12 bit so raw_capture should contain values from 0-1000.  If you remove the offset by subtracting 256 the raw_capture will contain values from 0-744 so I'd suggest compressing those 744 values into 256 rather than just clipping all the values above 256, throwing away 2/3rds of the dynamic range.

```
# Convert to to float 32 
raw_capture = raw_capture.astype(np.float32)
# Divide existing values by max value (744) to get a value from 0-1, then multiply this by 255 and convert the array to uint8
raw_capture = (raw_capture / max_pixel * 255).astype(np.uint8)
```

Let me know if I'm missing something here or if you have any questions!  I'm keen to get this improvement merged and I hope the combined remove offset + scale will perform even better than the remove offset + clip which already seems to perform better than just the scale :-)

c3n:
I actually tried several variations, like /2, /4, /8, /16, *1, *2 (via additive binning instead of averaging) and *4 (so that includes your suggestion) and out of all those the *1 performed best (but /2 was nearly the same). Throwing away the dynamic range shouldn't be a problem (but dividing to avoid clipping would be a problem if the full moon is in the picture since that would again effectively throw away all the lower bits). We don't need a pretty picture with full dynamic range, we just need to find the stars in the picture and it doesn't really matter if sirius is value 255 oder 3700. What we could try is mapping the values non-linearly (although I really don't think it'd perform any better)

brickbots:
Thanks @c3n!  Sorry for any confusion, but I was not referring to binning or geometrically combining the raw pixels, just the way we go from the 12 bit values returned by the camera sensor to an 8 bit value for the image.

The centroiding system and solver both rely on the relative brightness of different stars in the pattern matching they perform so it's important to preserve the relative values of the various stars in the image.  We are also planning on using the brightness of the stars in the image to calculate background flux, so this is another important reason for trying to preserve the relative pixels values.

The max pixel value used in the scaling is not dependent on the content of the image, just the expected range of values returned by the sensor, so won't change if the moon or other overly bright object is in frame.

Let me know if you have any questions and thank you again for your contribution!

c3n:
Yes I know what you mean, but I tried the /2 ... *4 thing and /4 is already much worse than /2 or /1 despite still requiring clipping. what it does is basically:
raw_capture = raw_capture / 4
while the max_value for HQ cam would be 2**12-256 = 3.840 (not 744, which would be 10bit), so more like
raw_capture = raw_capture / 15
and that is basically what it already does in the release code (the 256 offset divided by 15 is just a background value of 17 which doesn't change all that much in the resulting picture). If you want to I can rerun the tests over the weekend and paste the results here but the short version is: dividing by anything greater than 2 finds less stars

c3n:
I reran the tests on a bunch of images with 0.1s - 1.6s and analog gain 20-22 and here's the results:

`offset=256, maxval=255, sz=512: [('images', 45), ('matches', 187), ('maxstar', 255), ('solutions', 40), ('stars', 276), ('unmatched_stars', 89)]
offset=256, maxval=511, sz=512: [('images', 45), ('matches', 173), ('maxstar', 237), ('solutions', 39), ('stars', 265), ('unmatched_stars', 92)]
offset=256, maxval=511, sz=760: [('images', 45), ('matches', 358), ('maxstar', 255), ('solutions', 38), ('stars', 456), ('unmatched_stars', 98)]
offset=256, maxval=255, sz=760: [('images', 45), ('matches', 380), ('maxstar', 255), ('solutions', 37), ('stars', 468), ('unmatched_stars', 88)]
offset=256, maxval=1023, sz=760: [('images', 45), ('matches', 297), ('maxstar', 252), ('solutions', 37), ('stars', 404), ('unmatched_stars', 107)]
offset=0, maxval=511, sz=512: [('images', 45), ('matches', 158), ('maxstar', 255), ('solutions', 35), ('stars', 231), ('unmatched_stars', 73)]
offset=256, maxval=1023, sz=512: [('images', 45), ('matches', 116), ('maxstar', 188), ('solutions', 35), ('stars', 204), ('unmatched_stars', 88)]
offset=0, maxval=511, sz=760: [('images', 45), ('matches', 296), ('maxstar', 255), ('solutions', 34), ('stars', 389), ('unmatched_stars', 93)]
offset=256, maxval=2047, sz=760: [('images', 45), ('matches', 171), ('maxstar', 227), ('solutions', 34), ('stars', 287), ('unmatched_stars', 116)]
offset=0, maxval=1023, sz=512: [('images', 45), ('matches', 117), ('maxstar', 219), ('solutions', 34), ('stars', 194), ('unmatched_stars', 77)]
offset=0, maxval=1023, sz=760: [('images', 45), ('matches', 253), ('maxstar', 255), ('solutions', 33), ('stars', 337), ('unmatched_stars', 84)]
offset=256, maxval=2047, sz=512: [('images', 45), ('matches', 93), ('maxstar', 149), ('solutions', 33), ('stars', 155), ('unmatched_stars', 62)]
offset=0, maxval=2047, sz=760: [('images', 45), ('matches', 156), ('maxstar', 236), ('solutions', 31), ('stars', 238), ('unmatched_stars', 82)]
offset=256, maxval=4095, sz=760: [('images', 45), ('matches', 120), ('maxstar', 149), ('solutions', 31), ('stars', 191), ('unmatched_stars', 71)]
offset=0, maxval=2047, sz=512: [('images', 45), ('matches', 77), ('maxstar', 171), ('solutions', 31), ('stars', 131), ('unmatched_stars', 54)]
offset=256, maxval=4095, sz=512: [('images', 45), ('matches', 38), ('maxstar', 89), ('solutions', 26), ('stars', 95), ('unmatched_stars', 57)]
offset=0, maxval=4095, sz=760: [('images', 45), ('matches', 85), ('maxstar', 165), ('solutions', 25), ('stars', 143), ('unmatched_stars', 58)]
offset=0, maxval=4095, sz=512: [('images', 45), ('matches', 23), ('maxstar', 105), ('solutions', 23), ('stars', 73), ('unmatched_stars', 50)]
offset=0, maxval=255, sz=512: [('images', 45), ('unmatched_stars', 0)]
offset=0, maxval=255, sz=760: [('images', 45), ('unmatched_stars', 0)]`

As you can see setting the maxval to 255 or 511 (which is basically throwing away the top 4 or 3 bits) gives the best results. 
As for the problems with clipping I see two solutions:
1) Looking at the code the solver doesn't even get the relative brightness and for SQM measurements it would be easy to just return 2 pictures, one with the lower bits and one with either all the bits or the upper 8 bits and then use that for SQM estimation
2) After solving I could detect the brightest star in the image and adjust the "digital gain" so that it's <255

Let me know what you think!

PS: Here's the code I used to test the values:

`
import os
import glob
import tetra3
import numpy as np
from PIL import Image
from collections import Counter
#from picamera2 import Picamera2
#from matplotlib.pyplot import imshow
from tetra3 import cedar_detect_client

cedarbin="/Users/c3n/kram/cedar/cedar-detect/target/debug/cedar-detect-server"
npz="/Users/c3n/kram/cedar/cedar-solve/tetra3/data/default_database.npz"

cedar_detect = cedar_detect_client.CedarDetectClient(binary_path=cedarbin)
t3 = tetra3.Tetra3(npz)

def process_raw(raw, offset=256, maxval=4095, sz=512):
    r = {}
    scaled = (raw.astype(np.float32) - offset) * 255 / maxval
    clipped = np.clip(scaled, 0, 255).astype(np.uint8)
    img = Image.fromarray(clipped).resize((sz, sz))
    np_image = np.asarray(img, dtype=np.uint8)
    centroids = cedar_detect.extract_centroids(np_image, sigma=8, use_binned=True)
    if not centroids:
        return r
    maxstar = 0
    for x,y in centroids:
        for xx in range(2):
            for yy in range(2):
                maxstar = max(maxstar, np_image[int(x)+xx, int(y)+yy])
    r["stars"] = len(centroids)
    r["maxstar"] = maxstar
    solution = t3.solve_from_centroids(
                            centroids,
                            (sz, sz),
                            fov_estimate=12.0,
                            fov_max_error=4.0,
                            match_max_error=0.005,
                            # return_matches=True,
                            target_pixel=(sz//2, sz//2),
                            solve_timeout=1000
                        )
    r["matches"] = solution["Matches"] or 0
    return r

results = {}
for maxval in [4095, 2047, 1023, 511, 255]:
    for offset in [256, 0]:
        for sz in [512, 1520//2]:
            name = f"offset={offset}, maxval={maxval}, sz={sz}"
            results[name] = r = Counter()
            for fn in sorted(glob.glob("/Users/florianbusch/1/*/*.raw")):
                with open(fn, "rb") as f:
                    try:
                        raw = np.frombuffer(f.read(), dtype=np.uint16).reshape(1520, 1520)
                    except ValueError as e:
                        #print(f"Error reading {fn}: {e}")
                        continue
                thisr = process_raw(raw, offset=offset, maxval=maxval, sz=sz).items()
                r["images"] += 1
                if thisr:
                    r["solutions"] += 1
                    for k,v in thisr:
                        if k.startswith("max"):
                            r[k] = max(r[k], v)
                        else:
                            r[k] += v

for k,v in sorted(results.items(), key=lambda o:(o[1]["solutions"], o[1]["matches"]), reverse=True):
    v["unmatched_stars"] = v["stars"] - v["matches"]
    print(f"{k}: {sorted(v.items())}")
`

brickbots:
Thanks again @c3n for the diligent work here.  Couple of thoughts, and then a suggestion for wrapping this up:

- I was sloppy with my math in my example, sorry for that.  The max that should be used for the scale to preserve the relative values would be 4096 (max 12 bit value)- 256 (the floor you've determined).  So 3840.  I should never try to do base 2 in my head :-)  You didn't test this exact value+offset, but I figure the results would be between the 2048 and 4096 + 256 offset 
- It's not too surprising that some amount of digital gain, either by clipping (approximately 15x boost) or other means may yield more stars found and matched, which is a good thing.  There may be some point where the number of false centroids from noise slow things down, but Ceder_detect is both fast and effective.. so may not be an issue.  It would be interesting to see the same test results + solve/detect time just to make sure it's not getting pathological in there somewhere.  
- We used to use the max pixel value in the image to 'auto-scale' but this yielded a lot of variation which looked bad in the preview, made it hard to focus, and would not be acceptable for the SQM stuff.  So we went with the full range scale which is in there now.  I think avoiding the dynamic scale is a good thing, especially as we add auto-exposure in future mix.

So with all that in mind, I think adding a multiplier is probably a good thing for improving the solver performance, and the offset is a definite win with no downside.  Here's my suggestion:

* Add in an explicit multiplier and offset for each sensor.  This makes the code clear and readable and helps explain what is going on.  It also gives a universal target for either config or tuning over time.  The multiplier/offset can be set in the conditional per sensor next to the max_value.  While you are there, if you don't mind making the imx477 case explicit rather than the 'else' I would appreciate that.  I should have done that myself.
* From there, use a single code path for processing the cropped raw 16 bit image.  Break out the operations to make it really clear for other folks who may come along later:

 `# Apply sensor floor offset
 raw_capture = raw_capture - sensor_offset
 
 # Apply digital gain multiplier
raw_capture = raw_capture * multiplier

# Scale to 8 bit based on original sensor max value - offset
# Go to float for this calculation
raw_capture = raw_capture.astype(np.float32)

# Calculate ratio of pixel value to max value and multiply by 255
# to get 8-ish bit values (some will be higher than 255 due to digital gain multiplier)
raw_capture = raw_capture / (max_value - sensor_offset) * 255

# Clamp and convert to 8 bit
raw_capture = np.clip(raw_capture, 0, 255).astype(np.uint8)
`

This gets the benefit of the floor, the ability to set a specific digital gain multiplier, and is not quite as mysterious as applying a de-facto 15x  multiplier though the clamp 👍

c3n:
Updated it as per your specs. If you have some problematic RAWs somewhere of the other two sensor type I could run the tests to check if digital gain does anything

PR #336, opened 2025-07-31
Contrast reserve

WimDeMeester:
Adds the calculation of contrast reserve to PiFinder.
The value of the contrast reserve is shown in the object detail page:
2024-12-23 15_12_20-Luma core Display Emulator (piFinder)

Remarks:
- The contrast reserve is just a number... That is visible on the object detail page.  I have added an extra screen when you click on the square button that show more information about the contrast reserve.  I would like to keep it this way, because once you know what the contrast reserve means, that number is very meaningful.  

image

Some more information about the contrast reserve: 

| Contrast Reserve | Visibility             |
|------------------|------------------------|
| < -0.2           | Not visible            |
| -0.2 < CR < 0.1  | Questionable           |
| 0.1 < CR < 0.35  | Difficult              |
| 0.35 < CR < 0.5  | Quite difficult to see |
| 0.5 < CR < 1.0   | Easy to see            |
| 1.0 < CR         | Very easy to see       |

This pull request will solve #129


WimDeMeester:
@brickbots As far as I am concerned, this PR can go in the next version of PiFinder.

PR #338, opened 2025-08-20
Eq-mount support under IMU dead-reckoning tracking

TakKanekoGit:
Work-in-progress...

**Features:**
- IMU dead-reckoning tracking for EQ mounts.
- PiFinder no longer needs to be placed upright (it can be mounted as any angle as long as the camera points the same way as the scope).
- For altaz mounts, the charts are displayed in horizontal coordinates (Zenith-up) so that alt/az motion will be left/right and up/down. For EQ mounts, the charts are displayed in equatorial coordinates (NCP up) so that az/dec motion will be left/right and up/down.

**Changes:**
- `imu_pi.py`: IMU data coordinates are no longer altered here depending on the PiFinder type. IMU outputs will be they are from the IMU. This is now done in `imu_dead_reckoning.py`. 
- `integrator.py`: Substantial changes. Tracking is done in the equatorial coordinates rather than altaz coordinates. Use quaternion for the IMU dead-reckoning. Sets `solver["Roll"]` depending on the mount type for chart display.
- `pointing_model\`: The new classes and functions are grouped up here. It's possible that we could extend this in the future?
- `requirements_dev.txt`: Requires numpy.quaternion.

This PR touches the core of the IMU functionality so it still needs to be tested thoroughly on all PiFinder types and also on altaz and EQ mounts.

**Issues:**
- [x] Fix issues after merging with the auto exposure feature
- [ ] Chart rotates 180 degrees to the East of Mirfak (from sky testing): Pointings in RA, Dec were OK but in Chart mode, the chart rotated 180 degrees as I crossed to the East of around Mirfak in Perseus with the result that slewing to the 
East showed up as moving to the right.
- [ ] Southern hemisphere & equator EQ mode: Should it be South Pole up (roll = 180 deg)?
- [ ] Southern hemisphere & equator Alt/az mode: if lat < 0: Display Roll + 180 deg (south up).  Use North at equator

**Sky test IMU orientations**
- [x] Flat v2
- [ ] Left
- [ ] Right
- [ ] Straight
- [ ] Flat v3
- [ ] as_dream


python/PiFinder/pointing_model/README.md
```diff
@@ -0,0 +1,345 @@
...
+* On an altaz mount, we define +y as the vertical direction of the scope and +x
+  as the horizontal direction to the left when looking along the boresight.
+* In the ideal case, the Scope frame is assumed to be the same as the Gimbal
+  frame. In reality, there may be errors due to mounting or gravity.
+
+#### Camera frame
+* The camera frame describes the pointing of the PiFinder's camera. There will
+  be an offset between the camera and the scope.
+* $+z$ is the boresight of the camera, $+y$ and $+x$ are respectively the
+  vertical and horizontal (to the left) directions of the camera.
```
jscheidtmann: As we are assuming that the camera is pointing into the same direction as boresight (up to a small error). I was wondering, where do you take care of the possibility of a rotation between camera and scope frame?
Ok, got it: the `q_cam2scope`.

python/PiFinder/pointing_model/README.md
```diff
@@ -0,0 +1,345 @@
...
+frame
+
+### Alignment
+
+The alignment offset between the PiFinder camera frame and the scope frame is
+determined during alignment of the PiFinder with the scope and is assumed to be
+fixed. The goal of alignment is to determine the quaternion $q_{cam2scope}$.
+
+During alignment, the user user selects the target seen in the center the
+scope, which gives the (RA, Dec) of the scope pointing but not the roll. We can
```
jscheidtmann: ... center of the eyepiece, which ...

python/PiFinder/pointing_model/README.md
```diff
@@ -0,0 +1,345 @@
...
+Rearranging this gives,
+
+$$q_{cam2scope} = q_{eq2cam}^{-1} \; q_{eq2scope}$$
+
+Note that for unit quaternions, we can also use the conjugate $q^*$ instead of
+$q^{-1}$, because the conjugate is slightly faster to compute.
+
+Roll returned by plate-solving is not relevant for pointing and it can be
+arbitrary but it is needed for full three degrees-of-freedom dead-reckoning by
+the IMU.
```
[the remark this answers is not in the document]
TakKanekoGit: Yes, this could be possible but I think it's better to make it a manual setting because you could have weird cases for a Dob on an equatorial platform or a Portabll. I could imagine it could be annoying if it kept switching between the different modes.

python/PiFinder/pointing_model/README.md
TakKanekoGit: Fixed.

PR #343, opened 2025-09-08
exit on error, allow reruns, disable unwanted ModemManager

d33psky:
There should be 0 impact on existing use. Changes are:
- exit on error, so the script becomes reliable
- the script can now be run multiple times, it skips/adapts to parts already done
- disable the unused modem manager service, this frees up memory and improves battery life


brickbots:
Thanks @d33psky for this great improvement on a foundational piece!

PR #345, opened 2025-09-09
configurable GPS baud rate

d33psky:
The default (old) GPS units talk at 9600 baud
The newer UBlox-10 GPS series talk at 115200 baud
This PR adds a menu item to select between these 2 speeds and it updates the gpsd configuration in /etc/default/gpsd, and restarts the daemon on change, irrespective of whether UBLOX or classic GPSd is selected so that both will work.


d33psky:
Yeah I've thought of that as well.
Ideally we can easily detect all the various units but afaik we do not have that (yet).

The "GPS Type" menu lists
- GPSD (generic)
- UBlox
Both can have 9.6k and 115k2 devices.

We could extend it like this : 
- GPSD 9600 (generic)
- GPSD 115200 (generic)
- UBlox 9600
- UBlox 115200
But that looks messy too.

Maybe we should rename the "GPS Baud Rate" menu to "GPS chip" to enumerate the various GPS chipsets ?
- GT-U7
- Foxeer-M10Q
- GEP-M10

Or more generic like:
- UBlox-6
- UBlox-10

A chipset list will probably become too long.

mrosseel:
yeah it's not clear what the best approach is here, mainly concerned about users changing defaults and getting into trouble. If @brickbots is fine with your original proposal we can start there.

Further thoughts:
- There's mainly the V1 gps (gpsd 9600) and the V2/V2 gps (ublox 9600), the rest are custom gps'es. Would make it easy for users and still possible for custom gps'es to be activated

PR #348, opened 2025-09-19
Degraded Operation if IMU cannot be detected

jscheidtmann:
I had a strage thing last evening: The IMU was not detected and scanning the i2c bus didn't reveal any devices. That means the _init_ code in Imu did throw an Exception, which in turn means that the event loop process terminates, rendering the PiFinder unusable. 

As the IMU is not a critical piece for operation, it should fall back to imu_fake's implementation of Imu and display a warning instead. User can detect that there's no Imu in the status screen. 

This is a try for implementing that.


jscheidtmann:
@brickbots Ready for merge. Tested on an actual unit, whose IMU mysteriously works again (so I faked a problem by raising an exception in `_init_`)

brickbots:
Thank you @jscheidtmann!  This should be uncommon, but it's nice to know it's more visible and not a complete show stopper!

PR #379, opened 2026-02-12
NixOS support

mrosseel:
## Summary
- Full NixOS-based system for PiFinder (replaces Raspbian)
- Declarative system configuration via Nix flake
- SD card image, netboot, and migration bootstrap tarball builds
- Software update via nixos-rebuild with GitHub release/PR channels

## Test plan
- [ ] Flash SD image and verify boot
- [ ] Test WiFi AP and client mode switching
- [ ] Test software update UI channels
- [ ] Test hostname rename via web UI


nixos/networking.nix
```diff
@@ -0,0 +1,189 @@
+{ config, lib, pkgs, ... }:
+{
+  networking = {
+    hostName = "pifinder";
```
jscheidtmann, line 4: Does networking.nix hardcode the hostName and revert it every time nix is invoked? (Cf. Hashed password vs. initialPassword)

nixos/networking.nix
```diff
@@ -0,0 +1,189 @@
...
+      allowedTCPPorts = [ 80 ];     # PiFinder web UI (other ports via service openFirewall)
+    };
+  };
+
+  # Robust time sync for the RTC-less Pi: NTP= servers are always tried (and
+  # combined with any per-interface/DHCP servers), so a dead DHCP-advertised
+  # NTP server can't block the clock. FallbackNTP alone is skipped whenever a
+  # per-interface server is known — too fragile to rely on for first-boot
+  # migration, which gates the binary-cache fetch on a synchronized clock.
+  services.timesyncd.servers = [
```
jscheidtmann, line 20: Does timesyncd support sourcing from GPS time?

nixos/networking.nix
```diff
@@ -0,0 +1,189 @@
...
+      type=wifi
+      # Never self-start: NetworkManager would activate the AP instantly at
+      # boot (own radio, no scan needed) and win the race against a client
+      # network that still has to scan + associate, then stay on it. The AP is
+      # brought up only on demand by pifinder-wifi-fallback below.
+      autoconnect=false
+
+      [wifi]
+      mode=ap
+      ssid=PiFinderAP
```
jscheidtmann, line 62: How does PiFinder with nix handle changes of the WiFi AP name? Is this overwritten every time a new version is selected and downloaded?

nixos/networking.nix
jscheidtmann: It seems that the hostName is cared for later on. Please confirm

mrosseel: Hi, yes the hostname is stored in the pifinder data area

nixos/networking.nix
mrosseel: No idea, was this a feature we had before? If not I propose to fix that in a PR against this branch. It's already a big migration , want to wrap it up asap

PR #392, opened 2026-03-04
Lynga loader read for Pull Request

casazza:
Lynga Open Cluster Catalog loader.
Unit tests needed change to allow for 1 object at a Dec or exactly 0.0


casazza:
@mrosseel Sorry for the long delay. I have been getting the final push of the new version of TonightsSky through beta. I will get a chance to look at this later this week or weekend.

casazza:
@mrosseel identify_catalog_type serves a different purpose than ObjectFinder. ObjectFinder resolves whether an object already exists in the database by designation — it's a deduplication tool. identify_catalog_type classifies whether a name string is a PiFinder-recognized catalog prefix (NGC, IC, M, etc.) so it can be routed to aka_names for cross-matching versus common_names for the names table. These are two distinct operations. That said, if the project would prefer this classification list live in a shared location (e.g. a constant in catalog_import_utils.py) rather than repeated per loader, I'm happy to refactor it there — harris_loader.py has the same list and it could be consolidated.

mrosseel:
> @mrosseel identify_catalog_type serves a different purpose than ObjectFinder. ObjectFinder resolves whether an object already exists in the database by designation — it's a deduplication tool. identify_catalog_type classifies whether a name string is a PiFinder-recognized catalog prefix (NGC, IC, M, etc.) so it can be routed to aka_names for cross-matching versus common_names for the names table. These are two distinct operations. That said, if the project would prefer this classification list live in a shared location (e.g. a constant in catalog_import_utils.py) rather than repeated per loader, I'm happy to refactor it there — harris_loader.py has the same list and it could be consolidated.

hi, understood, maybe indeed refactor it out, no big deal but it's a bit cleaner

casazza:
@mrosseel is that a request to refactor this in the Lynga and Harris loader or the suggestion that this be consider acceptable technical debt that is put in the backlog?

mrosseel:
> @mrosseel is that a request to refactor this in the Lynga and Harris loader or the suggestion that this be consider acceptable technical debt that is put in the backlog?

I was indeed saying that a refactor would be good, but if you don't have the time for that now it's not really a blocking issue.

casazza:
@mrosseel Now is not a good time. I have a few other projects keeping my hands full. I will take this on when I do the quasar catalog loader.

mrosseel:
> @mrosseel Now is not a good time. I have a few other projects keeping my hands full. I will take this on when I do the quasar catalog loader.

ok no worries then, is it ready for testing?

casazza:
@mrosseel, Yes I think it is ready for testing.

brickbots:
Merging!  Thanks for the new catalog contribution!!!

PR #395, opened 2026-03-17
feat: add pygame keyboard input with intuitive key bindings

mrosseel:
## Summary
- When using pygame display (`-fh`), capture keyboard events directly in the main event loop instead of relying on PyHotKey/pynput (which can't access keyboard on Wayland)
- Intuitive key mapping: `+`/`=` and `-` for PLUS/MINUS, Enter/Space/Z for SQUARE, Ctrl+key for ALT_ variants (SQUARE+key on hardware)
- When pygame display is active with `--keyboard local`, automatically uses `keyboard_none` since the main loop handles key capture

## Test plan
- [ ] Run with `-fh --keyboard local -x` and verify +/- keys work for PLUS/MINUS
- [ ] Verify Enter/Space select items (SQUARE)
- [ ] Verify Ctrl+Plus and Ctrl+Minus trigger ALT_PLUS/ALT_MINUS
- [ ] Verify Ctrl+arrows trigger ALT_ arrow variants
- [ ] Verify arrow keys, number keys still work as before


mrosseel:
Closing to retarget against upstream main instead of upstream release

PR #397, opened 2026-03-17
feat: add pygame keyboard input with intuitive key bindings

mrosseel:
## Summary
- Add pygame-based keyboard input support for development/testing without hardware keypad
- Map intuitive key bindings (arrow keys, enter, escape, etc.) to PiFinder button events
- Enable with `--keyboard local` flag

## Test plan
- [x] Run with `--keyboard local --camera debug` and verify all key bindings work
- [ ] Verify no impact when using hardware keypad (default mode)


jscheidtmann:
Is that a display keyboard (that shows the keyboard binding in a UI)?

I created already some extensive keybindings, see the comment at the top of [keyboard_local.py](https://github.com/brickbots/PiFinder/blob/release/python/PiFinder/keyboard_local.py)

mrosseel:
> Is that a display keyboard (that shows the keyboard binding in a UI)?
> 
> I created already some extensive keybindings, see the comment at the top of [keyboard_local.py](https://github.com/brickbots/PiFinder/blob/release/python/PiFinder/keyboard_local.py)

hi, no it's not a visualisation, pyhotkey does not work on wayland so I figured a pygame approach to steer the pygame window might be cleaner.

I am working on some visualisation though, but that's not something expected to be checked in.

jscheidtmann:
Maybe this fixes my woes with website testing: pyhotkey on Mac delivers keypresses, even if the window is not focused. Hmm, need to look into a null display.

brickbots:
> Maybe this fixes my woes with website testing: pyhotkey on Mac delivers keypresses, even if the window is not focused. Hmm, need to look into a null display.

@jscheidtmann I've just checked in #435 which may be of interest to you!  Probably a good base for web testing and includes some shutdown logic

brickbots:
@mrosseel I've merged main into this branch and re-organized the initialization out of the main loop code.  I've tested it and its *very* nice to not have it pickup stray keystrokes when the PyGame window is not focused 🙌 

I'm ready to merge this if you are!

PR #399, opened 2026-03-20
fix: update observed filter cache after logging

mrosseel:
## Summary

- After logging an observation, the in-memory `observed_objects_cache` was not updated, so filtering by "observed/not observed" gave stale results within the same session
- Add `(catalog, sequence)` to the cache immediately after `log_object()` commits
- Set `obj.logged = True` on the `CompositeObject` so catalog filters see the change without a cache lookup

## Test plan

- [ ] Log an observation on an object
- [ ] Without restarting, filter catalog by "observed = Yes" — newly logged object should appear
- [ ] Filter by "observed = No" — newly logged object should no longer appear

### DIFF
--- a/python/PiFinder/db/observations_db.py
+++ b/python/PiFinder/db/observations_db.py
@@ -121,6 +121,10 @@ class ObservationsDatabase(Database):
         )
         self.conn.commit()
 
+        # Update cache so filters reflect the new observation immediately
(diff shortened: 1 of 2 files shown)


brickbots:
Thank you!  This is actually one of my pain points when I'm out observing and working through a list.

PR #402, opened 2026-03-24
feat: add manual date and location entry screens

mrosseel:
## Summary
- **UIDateEntry**: New date entry screen (YYYY-MM-DD) that automatically follows UITimeEntry when setting time manually, so users can set both time and date when GPS is unavailable
- **UILocationEntry**: New 3-step coordinate entry (lat → lon → alt) for manual location input, accessible from Tools > Place&Time > Set Location > Enter Coords
- **Location management**: "Load Location" shows saved locations (with empty state message), "Save Location" saves current position with a name prompt (shows "No location lock" if no fix)
- **Manual time protection**: Manual time/date is protected from GPS overwrite via `__datetime_manual` flag
- **Consistent navigation**: Right=confirm/forward, Left=back/cancel across all entry screens; last box stays selected instead of wrapping
- **Shared helper**: `Location.make_fix()` static method used by both UI callbacks and web server
- **GPS status**: Date now shown alongside time in detailed view
- **Separate resets**: "Reset Location" and "Reset Time/Date" are independent menu items

## Test plan
- [x] Navigate to Tools > Place&Time > Set Time/Date, enter time, confirm → date entry should appear pre-filled
- [x] Confirm date → verify datetime is set and persists (not overwritten by GPS)
- [x] Test Left key: goes back one box, cancels with popup on first box
- [x] Last box stays selected after filling (no wrap to first box)
- [x] Navigate to Tools > Place&Time > Set Location > Enter Coords
- [x] Enter lat → lon → alt flow, verify location is set
- [x] Test pre-fill from existing locked location
- [x] Tools > Place&Time > Set Location > Load Location shows "No locations" when empty
- [x] Tools > Place&Time > Set Location > Save Location saves current position, shows "No location lock" if no fix
- [x] Reset Location and Reset Time/Date work independently
- [x] GPS status detailed view shows date
- [x] Verify web interface location setting still works

### DIFF
--- a/python/PiFinder/main.py
+++ b/python/PiFinder/main.py
@@ -570,6 +570,7 @@ def main(
                             if (
                                 not location.source == "WEB"
                                 and not location.source.startswith("CONFIG:")
+                                and not location.source == "MANUAL"
                                 and (
                                     location.error_in_m == 0
                                     or float(gps_content["error_in_m"])
--- a/python/PiFinder/main.py
+++ b/python/PiFinder/main.py
@@ -611,12 +611,14 @@ def main(
                                         location.lon,
                                         location.altitude,
                                     )
-                        if gps_msg == "time":
+                        if gps_msg in ("time", "time_force"):
                             if isinstance(gps_content, datetime.datetime):
                                 gps_dt = gps_content
                             else:
                                 gps_dt = gps_content["time"]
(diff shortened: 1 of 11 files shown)


brickbots:
This feels like a, finally, complete and cohesive solution to the 'no gps' use-case.  Thanks @mrosseel !

PR #411, opened 2026-04-23
Telemetry recording and replay for IMU + plate-solve sessions

mrosseel:
## Summary
- Add telemetry recording: IMU samples (quat + raw gyro/accel), plate solves with predicted-vs-solved drift data, and target changes, written as compact JSONL sessions to `~/PiFinder_data/telemetry/`; optional per-solve frame capture via a new camera `save_image:` command. Location is kept in a separate sidecar file so sessions can be shared without leaking it.
- Add replay: sessions are converted back into `SolveResult`/`ImuSample` messages and fed through the integrator's normal apply/advance paths with original timing; the session's clock and location are force-applied for faithful Alt/Az and restored when replay ends. Start/stop from the new Telemetry menu (Record, Images, Load).
- Robustness: corrupt or truncated sessions are skipped line-wise instead of killing the integrator; IMU is captured from session start (also pre-first-solve), deduped on sample epoch; buffer overflow is logged; header datetime/location late-bind once GPS lock arrives.
- Tests: `test_telemetry.py` (record/replay round-trip equivalence) and `test_integrator_drift.py`.

## Test plan
- [x] `nox -s unit_tests` passes (includes new `test_telemetry.py` and `test_integrator_drift.py`)
- [x] `nox -s lint` passes
- [x] `nox -s type_hints` passes
- [x] Record a telemetry session on real hardware and verify replay produces matching pointing output


python/PiFinder/imu_pi.py
```diff
@@ -171,19 +171,27 @@ def imu_monitor(shared_state, console_queue, log_queue):
...
+        "gyro": None,  # Raw gyroscope angular velocity (rad/s)
+        "accel": None,  # Raw linear acceleration (m/s², gravity removed)
     }
 
     while True:
         imu.update()
         imu_data["status"] = imu.calibration
 
         # TODO: move_start and move_end don't seem to be used?
+        # Read raw sensor data for telemetry
```
TakKanekoGit: Wouldn't it make more sense to do this in imu.update() when the quaternions are read out?

python/PiFinder/imu_pi.py
```diff
@@ -171,19 +171,27 @@ def imu_monitor(shared_state, console_queue, log_queue):
...
+        try:
+            imu_data["gyro"] = imu.sensor.gyro
```
TakKanekoGit: The current version doesn't read gyro and accelerometer so we might get different behaviour, what with read/write, if we also start sampling additional data from the IMU. 

When I previously analysed the gyro recorded with an earlier version of `telemetry`, it wasn't very usable. That might be due to the BNO055 IMU because I also got poor performance in GYRONLY mode. 

Would it be an idea to make the default just to read in the quaternions and to have a configuration somewhere to enable reading in gyro and acceleration as an option?

TakKanekoGit (commented):
Hi! I just reviewed the part I'm familiar with (`imu_pi.py`). I use telemetry a lot so it'll be great to have it in `main`.

python/PiFinder/imu_pi.py
mrosseel: Good call — done in 3147e68f. `update()` now reads gyro/accel in the same gated pass as the quaternion (coherent sample, no per-spin I2C traffic) and stamps the read epoch, which the monitor uses as the sample timestamp. That also fixed a telemetry gap where stationary samples were never recorded. While in there I removed the stray `imu = Imu()` that made the ImuFake fallback unreachable.

python/PiFinder/imu_pi.py
mrosseel: Done in b1e9fed2 — raw gyro/accel capture is now opt-in via a `telemetry_raw_imu` config option, default off, so the default sampling path is quaternion-only as before. Given your findings on the BNO055 gyro quality, that seems like the right default.

PR #416, opened 2026-05-02
Imu fix chart orientation

TakKanekoGit:
This PR fixes chart display with the *quaternion* version of the integrator.

Fixed by rotating the chart by `-parallactic_angle` rather than `roll`.

When `mount_type` is alt/az, it will correctly display the chart in horizontal coordinates. If the `mount_type` is equatorial, it will display the chart in equatorial coordinates (north-up in norther hemisphere and south-up in southern hemisphere). If location or time is not available from the GPS, the chart display will default to equatorial coordinates (north-up).

Sky tested (f53401a) 1 May 2026 in the northern hemisphere.

### DIFF
--- a/python/PiFinder/calc_utils.py
+++ b/python/PiFinder/calc_utils.py
@@ -388,6 +388,25 @@ class Skyfield_utils:
 
         return ha_hrs * 180 / 12  # Hour angle [deg]
 
+    def radec_to_pa(self, ra_deg, dec_deg, dt):
+        """
+        Returns the parallactic angle of an object at (ra, dec) as seen from an
+        observer at latitiude, lat and time, dt. See hadec_to_pa() for how
+        parallactic angle is defined.
+
+        INPUTS:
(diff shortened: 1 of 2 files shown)


brickbots:
Hi @TakKanekoGit thanks for this great double-pr update for the chart screen.  I merged the later one firstand ended up with a conflict on this earlier one 😅   I *think* I've fixed it up and it seems to run fine, but before I merge this to main would you take a quick scan of integrator.py to make sure I didn't make a more subtle mistake?

TakKanekoGit:
Hi @brickbots -- Thank you! Actually, this PR (#416) is a direct descendent of the PR you merged in (#417) so all the functionalities from this PR is already in `main`.  Sorry my request wasn't clear.

Maybe the easiest option is to abandon this PR? If you're ok with that, I will remove it.

brickbots:
Hi Hi!  Yes, that sounds great, if #417 does the job then I'll close this.  Thanks again for this great update!

PR #459, opened 2026-06-09
Polar alignment for equatorial platforms & mounts

mrosseel:
## Polar alignment for equatorial platforms

Adds a **Polar Align** tool that helps you set the polar-axis tilt of an equatorial (Poncet-style) platform using the PiFinder's plate solver. You solve a few points while rotating the platform between them; the tool fits the platform's rotation axis, compares it to the celestial pole, and gives live push-to guidance for the altitude/azimuth adjusters until the axis is aligned.

Lives under **Tools → Experimental → Polar Align**.

### How it works (user flow)

1. **Intro** – pick the mode, wait for a GPS fix.
2. **Aim** – point away from the pole, let the camera solve, capture 2–3 points, rotating the platform ≥10° between each.
3. **Capture / wait** – holds while the camera solves each point.
4. **Adjust** – live correction target; shows how far (and which way) to turn the alt/az knobs, with a fit-quality verdict.
5. **Stats** – read-only detail of the last solve (fit, dAlt/dAz, axis RA/Dec, capture timing), available on demand from 2+ points.

A long-press marking menu offers redo-point, an RA/Dec-only roll toggle (for after a camera flop), and the stats view.

01-intro
02-aim
03-marking-menu
04-adjust
05-stats

### Solver — `PiFinder/polar_alignment/`

- New `polar_alignment` **package** (was a single module): the pipeline of `get_platform_adjustments` is split into named stages (`_epoch_align_solves`, `_two_solve_axis`, `_ignore_roll_axis`), and the physical knob composition is extracted from `correction_target` into `_altaz_knob_rotation`.
- **Annual aberration nudge** is applied on the `_altaz_knob_rotation` seam, using `skyfield.constants.C` and the shared `calc_utils.sf_utils` ephemeris (no module-local Loader / de421 load).
- Below-`MIN_SWEEP` early return uses the same sweep-sign convention as the full result.
- A seeded Monte-Carlo benchmark lives at `python -m PiFinder.polar_alignment.benchmark` (moved out of the importable module).

### UI

- `PiFinder/ui/polar_align.py` – the screens above.
- Shared push-to pointing rendering extracted to `PiFinder/ui/ui_utils.py` and reused by object details / list, so the platform-adjust arrows and the existing locate arrows share one implementation.
- Menu: Polar Align moved to **Tools → Experimental**; the old Align (Day) entry removed.

### Internationalisation

- Full **de / es / fr / zh** translations for the polar-alignment UI (intro, aim, capture, adjust, stats, marking menu, status messages).
- Technical labels kept verbatim per house convention (`Alt`, `Az`, `RA/Dec`, `EQ`, `GPS`); hint-bar strings use named placeholders for the button glyphs so they stay translatable.
- AI-assisted draft translations are tagged `# AI-TRANSLATED` in the `.po` files for human review; `.mo` catalogs recompiled.

### Tests

- `python/tests/test_polar_alignment.py` – axis-fit and `correction_target` knob-consistency sanity checks, including the documented apparent-vs-geometric pole distinction when `observation_jyear` is set; near-pole tolerances are commented (degrees→arcsin conditioning).

### Misc

- `pandas 1.5.3 → 2.0.3` (clears the `np.find_common_type` deprecation warning).
- Docs: user-guide menu map + quick-start updated for the new tool.

---

**Diff:** 24 files, ~+4.8k/−1.2k. Full smoke + unit suite passing; ruff and mypy clean.

Incorporates the earlier fork PRs #7 (annual aberration target) and #8 (correction-target tests), plus the #11 refactor/aberration-restore/experimental-menu batch.


python/PiFinder/polar_alignment/__init__.py
```diff
@@ -0,0 +1,968 @@
+"""
```
TakKanekoGit, line 1: Before making any changes, can I suggest asking Claude/Copilot to help re-architecture and restructure and refactor polar_alignment.py?

- I assume that the tests in place will help you to ensure that the behaviour remains the same after refactoring?
- Before implementing any code changes, I suggest discussing with the proposals with Mike to seek his sage advice.

Some requirements might be (which you might want to put in the prompts):

- Separate out the core algorithm that will be run by the PiFinder and the simulation/development code.
  - For example, you could create a `polar_alignment\ directory` to keep the files together.
  - `polar_alignment.py` for the core algorithms. 
  - `polar_alignment_simulations.py` for simulations. 
  - Another file for development scripts to run the simulations that are currently in the main block.
- The main algorithms are in `get_platform_adjustments()` but it's long with multiple algorithms. I propose to split each algorithm (two-attitude solve, three-vector solve, three-attitude solve) into separate functions.
  - The first small step might be to refactor `get_platform_adjustments()` into individual components and write tests for them before the big restructuring?

python/PiFinder/polar_alignment/__init__.py
```diff
@@ -0,0 +1,968 @@
...
+
+    lat = np.radians(latitude_deg)
+    dec = np.radians(dec_ax)
+    ha = np.radians(ha_ax)
+
+    # Standard equatorial → horizontal transformation
+    sin_alt = np.sin(lat) * np.sin(dec) + np.cos(lat) * np.cos(dec) * np.cos(ha)
+    alt = np.degrees(np.arcsin(np.clip(sin_alt, -1.0, 1.0)))
+
+    az = np.degrees(
```
TakKanekoGit, line 162: Can I recommend sticking to radians and SI units throughout and changing to degrees just at the interface, if necessary. Best practice in scientific computing is to use SI because unit conversion risks introducing bugs.

A nicer interface might be to use the RaDecRoll class. I think we're starting to use this in the PiFinder state interfaces? @brickbots can advise on how to best fit it in with the PiFinder architecture because I think he has a vision that he's moving towards.

python/PiFinder/polar_alignment/__init__.py
```diff
@@ -0,0 +1,968 @@
...
+                                solve, in degrees (0–360).
+                                To convert from hours: lst_deg = lst_hours * 15.
+    sigma_ra          : float   Expected 1-sigma RA noise in degrees (default 1').
+                                Used only when three or more solves are given.
+    sigma_dec         : float   Expected 1-sigma Dec noise in degrees (default 1').
+                                Used only when three or more solves are given.
+    sigma_roll        : float   Expected 1-sigma roll noise in degrees
+                                (default 1/R where R = field radius in radians;
+                                for a 10° field, R = 5° = 0.0873 rad giving
+                                sigma_roll = 11.4592 x sigma_RA/Dec ≈ 11.46x).
```
TakKanekoGit, line 676:
```
for a 10° field, R = 5° = 0.0873 rad giving sigma_roll = 11.4592 x sigma_RA/Dec ≈ 11.46x).
```

Is this right? The equation would imply that `sigma_roll` has "dimension" 1/radians. While radians are technically dimensionless, it might be worth checking?

python/PiFinder/polar_alignment/__init__.py
```diff
@@ -0,0 +1,968 @@
...
+                                Set higher to downweight roll relative to RA/Dec.
```
TakKanekoGit, line 677: If we assume that roll is ~10 times worse than RA/Dec error, does including roll help? I know you shared some Monte Carlo results but I couldn't follow it.

python/PiFinder/polar_alignment/__init__.py
```diff
@@ -0,0 +1,968 @@
...
+    Returns (axis, final_cost) where final_cost is the cost function value at
+    the solution, ready to use as fit_quality = sqrt(final_cost / n_obs).
+    """
+    x0_pos, x0_neg, x0_xp = _three_solve_initial_params(obs, axis_seed)
+    seeds = [x0_pos, x0_neg, x0_xp]
+    costs = [_three_solve_cost(x, obs, sigma_ra, sigma_dec, sigma_roll) for x in seeds]
+    best_seed_idx = int(np.argmin(costs))
+    x0 = seeds[best_seed_idx]
+
+    result = minimize(
```
TakKanekoGit, line 487: I have some questions and comments specifically on the iterative approach used here:

- Am I right in understanding that this uses all N (RA, Dec, Roll) solves? 
- There are 7 unknowns. I think the computational complexity of BFGS scales as the square of the number of unknowns. We want to solve for the RA axis of the mount, which only has 2 degrees-of-freedom. Do we need all these unknowns? 
- `three_solve_cost` minimises a scalar cost function. If you want to go the iterative optimisation approach, convergence is usually more efficient if the cost function returns a vector residual which gives the optimisation algorithm information about which direction to try next. `scipy.optimize.least_squares` can do this.

python/PiFinder/polar_alignment/__init__.py
```diff
@@ -0,0 +1,968 @@
...
+        t1_xp = _signed_sweep_angle(n_xp, b1, b2)
+        t2_xp = _signed_sweep_angle(n_xp, b2, b3)
+        x0_xp = [float(n_xp[0]), float(n_xp[1]), t1_xp, t2_xp, ra0, dec0, roll0]
+    else:
+        x0_xp = x0_pos  # degenerate: fall back to two-solve seed
+
+    return x0_pos, x0_neg, x0_xp
+
+
+def _refine_axis_three_solves(obs, axis_seed, sigma_ra, sigma_dec, sigma_roll):
```
TakKanekoGit, line 468: I'd like to open up a discussion about whether the iterative optimisation approach is the right one here.

1) I solved this problem using linear algebra because it's an Eigenvector problem. That seems to be the simpler and more efficient approach?

2) Am I right in understanding that you've taken an optimisation approach so that each of RA, Dec, Roll can be weighted? I wonder if that's necessary? 

The raw measurements originally in `tetra3` are rotation matrices. I can't quite picture what the uncertainty on these rotation matrices might look like but it's probably all mixed around so it may not be necessary to worry about weights if you use the rotation matrices directly as measurements? 

When you decompose the rotation matrix to (RA, Dec, Roll), it presumably projects a lot of the uncertainty along the Roll component so you need to worry about weights but if you're directly using rotation matrices, maybe you don't need to? It's just a hypothesis but it could probably be tested using Monte Carlo?

3) If it helps, for reference, this is my function to solve from N points:

```python
def estimate_axis(q_meas):
    """
    Estimate the axis of rotation from N-point measurements
    """
    if len(q_meas) < 2:
        raise ValueError("Need at least 2 measurements")
    
    # Construct the matrix equation: [R - I] = 0
    n_meas = len(q_meas)
    A = np.zeros((n_meas * 3, 3))
    for imeas, q in enumerate(q_meas):
        # TODO: Ensure quaternion continuity
        A[imeas * 3:(imeas + 1) * 3, :] = quaternion.as_rotation_matrix(q) - np.eye(3)

    # Solve for the RA axis
    _, sing_values, Vt = np.linalg.svd(A, full_matrices=True)
    ra_axis = Vt[-1, :]  # Axis of RA axis (associated with smallest singular value)
    print("sing_values ratio = ", sing_values[-2] / sing_values[-1])
    
    return ra_axis
```

python/PiFinder/polar_alignment.py
```diff
@@ -0,0 +1,968 @@
...
+
+
+# ── Monte Carlo benchmark ─────────────────────────────────────────────────────
+# Exploratory accuracy comparison of the two-solve, three-solve-optimised, and
+# RA/Dec-only (cross-product) methods across a range of pointing geometries and
+# sweeps.  This is not a test — the pass/fail correctness checks live in
+# tests/test_polar_alignment.py.  Run with `python -m PiFinder.polar_alignment`
+# to print the comparison table.
+
+if __name__ == "__main__":
```
TakKanekoGit: I propose to keep the main block to a minimal example. 
Please see my previous comment about moving some of the code in the main block to a development script.

python/PiFinder/polar_alignment.py
```diff
@@ -0,0 +1,968 @@
...
+            )
+            _t1 = _signed_sweep_angle(_n_ir, _bs[0], _bs[1])
+            _t2 = _signed_sweep_angle(_n_ir, _bs[1], _bs[2])
+
+            try:
+                _ps = _predict_three_solves(
+                    _n_ir, _t1, _t2, _obs_ir[0][0], _obs_ir[0][1], _obs_ir[0][2]
+                )
+                _ignore_roll_final_cost = sum(
+                    (_wrap180(_p[2] - _m[2]) / sigma_roll) ** 2
```
TakKanekoGit: If it's ignoring roll, why is it scaled by `sigma_roll`?

python/PiFinder/polar_alignment.py
```diff
@@ -0,0 +1,968 @@
...
+    #    Uses the cost at the actual solution parameters — no sweep re-estimation.
+    #    For ignore_roll=True: roll-only cost at the best-fit sweep for the
+    #    cross-product axis — measures roll consistency with the RA/Dec-derived
+    #    axis, with expected value ~1 under pure noise.
+    fit_quality = float("nan")
+    if len(solves) >= 3:
+        if ignore_roll and np.isfinite(_ignore_roll_final_cost):
+            fit_quality = np.sqrt(_ignore_roll_final_cost / len(solves))
+        elif not ignore_roll:
+            fit_quality = np.sqrt(_three_solve_final_cost / len(solves))
```
TakKanekoGit: **Fit quality calculation**
If I understand it correctly, `_three_solve_final_cost` is the weighted RMSE, which then makes fit_quality something almost like a reduced chi-square? If so, it might make more sense to use the reduced chi-square, which is a more conventional metric that's interpretable?

python/PiFinder/polar_alignment.py
```diff
@@ -0,0 +1,968 @@
...
+    _three_solve_final_cost = float("nan")
+    _ignore_roll_final_cost = float("nan")
+    if len(solves) >= 3:
+        if ignore_roll:
+            # RA/Dec-only: find the axis as the pole of the arc defined by
+            # the three boresight directions.  Roll is not used at all,
+            # making this immune to systematic roll errors (e.g. camera flop).
+            # Uses the first, middle, and last solve for the best arc coverage.
+            mid = len(solves) // 2
+            b1 = _boresight_vec(solves[0][0], solves[0][1])
```
TakKanekoGit: The `ignore_roll` option only uses 3 measurements (one at both ends and one in the middle). You can generalise this to any N >= 3 measurements.

For reference, here's what I did (I haven't validated this yet so that still needs to be done):

```python
def construct_measurement_matrix(samples: list # List of (timestamp, RaDecRoll)
                                 ):
    """ Construct the measurement matrix (measurements along rows) """
    measurements = np.zeros((len(samples), 3))
    for ii, (_, coord) in enumerate(samples):
        measurements[ii, :] = ra_dec_to_vec(coord.ra, coord.dec)
    return measurements

def solve_axis(
        measurements: np.ndarray  # (n_meas, 3) matrix of measurements along rows
        ) -> tuple[np.ndarray, np.ndarray]:
    """ 
    Solve for the RA axis of the mount (see README for the theory).
    measurements should be pre-selected. Needs at least 3 measurements.
    
    INPUTS:
    measurements: 
    - If the measurements are provided in equatorial coordinates, the ra_axis
    will be in equatorial coordinates.
    - Similarly, if the measurements are provided in horizontal coordinates, 
    the ra_axis will also be in horizontal coordinates.

    RETURNS:
    ra_axis: (3,) vector of RA axis
    sing_values: (n_meas,) vector of singular values for diagnostics (in descending order)
    """
    n_meas = measurements.shape[0]
    if n_meas < 3:
        return None, None
    
    # Create a matrix P that cyclically shifts the rows of measurements one down
    P = np.roll(np.eye(n_meas), 1, axis=1)
    # Solve for the RA axis
    _, sing_values, Vt = np.linalg.svd((P - np.eye(n_meas)) @ measurements, full_matrices=True)
    ra_axis = Vt[-1, :]  # Axis of RA axis (associated with smallest singular value)
    
    return ra_axis, sing_values  # ra_axis
```

python/PiFinder/polar_alignment/__init__.py
```diff
@@ -0,0 +1,968 @@
...
+    #    cross-product axis — measures roll consistency with the RA/Dec-derived
+    #    axis, with expected value ~1 under pure noise.
+    fit_quality = float("nan")
+    if len(solves) >= 3:
+        if ignore_roll and np.isfinite(_ignore_roll_final_cost):
+            fit_quality = np.sqrt(_ignore_roll_final_cost / len(solves))
+        elif not ignore_roll:
+            fit_quality = np.sqrt(_three_solve_final_cost / len(solves))
+
+    return dAlt, dAz, -sweep_deg, axis_ra, axis_dec, fit_quality
```
TakKanekoGit, line 804: **Fit quality**
I think what we ideally want is an uncertainty of the solution in angles in (RA, Dec) or just a single angle. 

If the angular uncertainty smaller than the angular offset from the pole, the user should attempt to adjust the mount. If the uncertainty is of the order of the angular offset, it's at the limit of accuracy and no further adjustment needed.

I wonder if that's possible?

Fit quality is only evaluated for the N>=3 case because N=2 is exactly solved. We could probably estimate the uncertainty in a principled way by calculating the Jacobians but it might be a bit of a hassle. By the way, `tetra3` can return the RMSE of the solve in arcsec, which might help.

Perhaps this can be an improvement for the future and not worth worrying about now?

python/PiFinder/polar_alignment/__init__.py
```diff
@@ -0,0 +1,968 @@
...
+  Column 2 of M is the boresight unit vector pointing at (RA, Dec):
+    M[:,2] = [cos(Dec)cos(RA),  cos(Dec)sin(RA),  sin(Dec)]
+  Construction:  M = Rz(RA) · Ry(90° − Dec) · Rz(−roll)
+  Extraction via direct element read-off — no gimbal-lock issues.
+
+Why LST is required
+-------------------
+The rotation axis extracted from the two plate solves is a direction in the
+equatorial frame, characterised by (Dec_axis, RA_axis).  Its angular distance
+from the NCP (= 90° − Dec_axis) is the total misalignment magnitude.  But to
```
TakKanekoGit, line 25: Does the algorithm support cases in the Southern hemisphere and near the equator (which pole do they even choose at the equator?).

TakKanekoGit (commented):
Wow. What a lot of work in such a short time 😃. I can't wait to test it under the stars. I've mainly focused on the algorithmic side for my review.

python/PiFinder/polar_alignment/__init__.py
alexis-cousein: It is correct -- and the reason why the PiFinder's large field makes such a difference.

python/PiFinder/polar_alignment/__init__.py
alexis-cousein: Yes, including roll helps...as long as the cost function used in the optimiser reflects the relative variance of noise due to limited precision.

That's because close to the pole roll noise is large but so is roll, whereas Ra/Dec changes are small.

The rather bizarre result is that for a PiFinder pointing to the pole is still better than pointing far from the pole and using only RA/Dec.

But for something with 1° field the variance in roll error would actually be 115 times worse than those in Ra and Dec. The algorithm would naturally degenerate into "take the Ra/Dec based only estimate as the seed, and mostly ignore roll errors". In other words, since the solution using three RA/dec is constrained,  the optimiser would do nothing. And the fit quality would be appalling close to the pole too. 

With such a setup, with a smaller field, if you have two points, just do a 2 point solve (it will be 'bad'), and if you have three, just fit the Ra/Dec on a circle (with three points). No surprises...that's what all the algorithms for polar alignment with astrocameras (and not PiFinders) actually do.

python/PiFinder/polar_alignment/__init__.py
alexis-cousein: Not N, just three. No one is going to add a fourth point, it's a PITA and it won't really make a material difference (yeah, 20 points would, but who is going to do *that*?)

Yes, we really want all these unknowns (but we want good seeds). The point is that we do not know, for the three solves, what the individual sweeps are. Changes to Ra/Dec/roll can be due to noise but also due to little differences in sweep. At least this is a formulation that works.

As for a vector residual: that's what the original SVD-based optimiser used, and it was just appalling. Once you go into a vector space you comingle RA/Dec/roll and there's no way to accurately model the noise differences you got from the solver. I went *back* to this to make it behave decently, degenerating properly into "use only RA/Dec" with large sweeps at low declination. Yes, a scalar cost function has its drawbacks but at least we're sure it's correct and roll errors don't infect it once you model it well.

This a) works b) is stable c) converges rapidly enough so that it'll be fast enough on a Pi d) degenerates into the known good algorithms at low Dec and large sweeps. Good enough for me ;-).

Of course if you find ways to improve it....that's why there's a small Monte-Carlo simulation, if you want to try something else.

python/PiFinder/polar_alignment.py
alexis-cousein: It ignores roll in _coming up with the best axis_, but not in the **fit quality**, which will express whether we have roll errors that seem inconsistent with what we expect.

Mind you: a bad fit quality will tend to express not that our noise model is wrong (unlikely) but more likely  if you try an alignment where pure RA/Dec works well (if you have a classical EQ mount), it will usually tell us whether we have real camera roll flop if we set ignore_roll to True.

Closer to the pole or with too small sweep angles (or worse, both), the fit quality will just tell you that roll_ignore=True does not work well.

python/PiFinder/polar_alignment.py
alexis-cousein: It's not exactly chi-square. And this metric is, I think, good enough.

python/PiFinder/polar_alignment.py
alexis-cousein: As I said: I doubt anyone would do four or five points, it'll only have limited impact and will be fussy (hey, some people prefer *two* points to three points even when it's less precise). But yes, it *is* actually possible.

python/PiFinder/polar_alignment/__init__.py
alexis-cousein: Exactly. Let's fix things that are broken, we'll know better where the bottlenecks are once we actually use this first.

python/PiFinder/polar_alignment/__init__.py
alexis-cousein: Yes, that's all taken care of. It DOES suppose that at positive latitudes you'll put what you labeled the "pole end" of the platform or the RA axis to what you think is the N celestial pole and negative latitudes point it to the SCP.

If you're close to the equator, just point the mount to the "correct" pole. You're not going to be using an alignment routine close to the pole as we do at 51°N, I think ;). 

The alignment routine will actually *MAKE* you point the axis at the pole we think is the right one. If you need to flip AZ 180°, you know the PiFinder thinks you're in the other hemisphere ;-).

The equator truly is degenerate, BTW, since the RA axis really is horizontal. You *could* actually pick either pole, but we'll pick a hemisphere based on the latitude from your GPS.

python/PiFinder/polar_alignment/__init__.py
alexis-cousein: Fair point, and if it all works well we could refactor it to make it cleaner. I must say I find degrees a little easier on the soul while debugging though, I have a hard time visualising Ra and Dec in radians.

python/PiFinder/polar_alignment/__init__.py
alexis-cousein: This is actually what Claude Fable 5 came up with now ;-). It tends to write a lot of utility functions that get reused left and right.

But yes, for the PiFinder you could just keep the solver and toss the sanity checks. But keeping them doesn't harm and lets you run just the .py file if you want to change the algorithm.

Of course yes, you could not have a main at all in one file and then have another for the sanity checks that includes it.

TakKanekoGit: I think we're all trying to get our heads round how to use AI for programming 😀.

There's a method to the madness of using AI for programming :-). 

If you just let the AI go, it does tend to generate lots of functions without much structure. You can use AI to help with refactoring. I think it's a worthwhile investment to restructure to help with maintainability and debugging issues in the future. The comments above outline possible ways to prompt it.

Of all the comments, I think this is my top priority item because it lays the groundwork for further development. @mrosseel is the master for this kind of stuff.

I'll try to reply to the other comments later...

python/PiFinder/polar_alignment/__init__.py
TakKanekoGit: Indeed. What you typically might do in that case is to just convert to degrees in the print statement.

python/PiFinder/polar_alignment/__init__.py
TakKanekoGit: Intuitively it makes sense that roll uncertainty might be ~1/R but I'd like to understand the derivation to check that we're using the right factor...

The factor 1/R suggests that the roll is determined by the outer stars at angle R from the centre but presumably the stars less than R contribute to reducing the uncertainty so maybe the factor could be less than 1/R?

python/PiFinder/polar_alignment/__init__.py
TakKanekoGit: Interesting. Can I suggest you write up your knowledge you mention here and in Discord? Perhaps at the top of this file or in a separate README? I think it's important to capture your knowledge that's not in the code somewhere.

TakKanekoGit: Rather than trying to support all scenarios, would a simpler approach be to allow the user to do the swing at a Dec within some allowed range from the pole? @mrosseel -- could that, perhaps, be part of the UI flow?

> That's because close to the pole roll noise is large but so is roll, whereas Ra/Dec changes are small.

If the roll error gets worse near the pole, wouldn't you expect `sigma_roll` in `_three_solve_cost()` to be dependent on the angle from the pole rather than a fixed value?

python/PiFinder/polar_alignment/__init__.py
TakKanekoGit: I agree that no one's going to do more than 2 points if you don't have to. 

My back-of-the-envelope calculation also suggests that improvements from additional measurements is pretty minimal. If so, the 3-point method might not give a big improvement over the 2-point method?

I'm still not convinced you need to solve for all 7 unknowns. Do you need 7 unknowns because you want to calculate the goodness of fit? Maybe there's a better way? Can you explain how your original SVD-based solution worked?

python/PiFinder/polar_alignment.py
TakKanekoGit: Hmmm. I'm still not convinced by this. 

If I understand the code, if `ignore roll==True`, it finds the RA-axis from the (RA, Dec) measurements. Then calculates the `fit_quality` by predicting the expected roll and comparing with the roll from the measurements. I'm not sure it's a fair comparison to compare this to a model fitted to the (RA, Dec, Roll) measurements?

python/PiFinder/polar_alignment.py
TakKanekoGit: Ok - resolved.

python/PiFinder/polar_alignment.py
TakKanekoGit: I agree that people probably aren't going to do many points in practice. 

I think that the code is cleaner if you can generalise it than the specific case.

python/PiFinder/polar_alignment/__init__.py
TakKanekoGit: I've created issue #463. We can close this.

python/PiFinder/polar_alignment/__init__.py
TakKanekoGit: That's nice. Using location to automatically determine NCP or SCP sounds like the way to go. 

For people near the equator, maybe a user setting might help? @mrosseel is probably best placed to advise on this? 

For reference, for charts, there's a similar option:   `config_object.get_option("chart_coord_sys")` with values like "horiz", "eq_auto", "eq_north_up" and "eq_south_up". "eq_auto" chooses North/South up depending on location.

python/PiFinder/polar_alignment/__init__.py
alexis-cousein: No, the roll error does not get worse. The *roll* gets larger so the *relative* signal to noise gets better because the roll noise does not change. I did not say the pole roll noise was large*r*, just large with respect to RA and Dec noise.

python/PiFinder/polar_alignment/__init__.py
alexis-cousein: The 3 point method *is* better than the 2-point method everywhere, and if you do not have access to the pole the 2 point method may do poorly (because of the noise on roll). 2 points also do not give you a fit_quality.

python/PiFinder/polar_alignment.py
alexis-cousein: It *is* indeed not directly comparable *to the other fit quality criterion*, but it does give you an indication of whether the RA/Dec fit is reliable or might be off (or alternatively, if you know it's not off, tell you something about systematic roll errors, like camera roll flop).

python/PiFinder/polar_alignment/__init__.py
alexis-cousein: Personally, I would neither overload the chart_coord_sys (there are other reasons for wanting the chart in a different orientation than latitude) nor introduce yet another tunable that no one has ever requested or is likely to ever want. But I'll let @mrosseel decide.

python/PiFinder/polar_alignment/__init__.py
TakKanekoGit: Ok - I think understand now. So it's correct to use the fixed `sigma_roll`? If so, we can close this. 

Would be still good to write up the subtleties 😀

python/PiFinder/polar_alignment/__init__.py
TakKanekoGit: Ok - Let's see how the 2 vs 3-point methods perform in real life.

Please have a think about whether 7 unknowns is necessary. I'd also be interested in your SVD-based solution.

python/PiFinder/polar_alignment.py
TakKanekoGit: Ok - let's go with it.  You can close this one.

TakKanekoGit:
I set the review as only comments and not change requests so they don't have to be resolved. I'll leave it to @alexis-cousein and @mrosseel to decide what you want to do with them. Good luck!

PR #465, opened 2026-06-11
fix(solver): clear leaked cedar_detect shmem segment on startup

mrosseel:
A solver process that is killed (or crashes) never runs the cedar client's close(), leaving the POSIX shared-memory segment /cedar_detect_image in /dev/shm. The next run's _alloc_shmem() then calls SharedMemory(create=True) on the same name and dies with FileExistsError on every solve, so plate solving never recovers until the segment is removed by hand.

PFCedarDetectClient now unlinks any pre-existing segment when constructed, making solver restarts self-healing. Safe because PiFinder runs a single solver process, so any existing segment is necessarily stale.

==> I've seen this mysterious error on various dev runs and 'in-the-field', and if it happens in the field your session is over. Seems like a safe fix.

### DIFF
--- a/python/PiFinder/main.py
+++ b/python/PiFinder/main.py
@@ -370,6 +370,12 @@ def main(
     integrator_logqueque: Queue = log_helper.get_queue()
     imu_logqueue: Queue = log_helper.get_queue()
 
+    # Refuse to start if another instance is already running. A second copy
+    # would otherwise boot and let its subsystems (web/pos-server ports, cedar
+    # shmem, hardware devices) silently collide with the live one.
+    if not utils.acquire_single_instance_lock():
+        return
+
     # Start log consolidation process first.
     log_helper.start()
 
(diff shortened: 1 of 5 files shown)


jscheidtmann:
I thought @brickbots moved the cedar process into a separate service. If that doesn‘t startup, it should clean the shmem, not PiFinder.

mrosseel:
> I thought @brickbots moved the cedar process into a separate service. If that doesn‘t startup, it should clean the shmem, not PiFinder.

I indeed remember him doing this, but I had the error on my local machine so either the service cleanup is not working or the service is not running. I'll check

jscheidtmann:
If you‘re talking dev machine, that‘s different and handling of that case (file cannot be created, because it exists) is different by OS.

mrosseel:
> I thought @brickbots moved the cedar process into a separate service. If that doesn‘t startup, it should clean the shmem, not PiFinder.

I checked a few days ago and the creation of the shmem file is outside of the service

PR #470, opened 2026-06-15
Vectorize comet propagation to fix locked-on-target CPU hog

brickbots:
> **⚠️ Attribution update.** This PR's original write-up framed the comet CPU hog as the cause of the 2.6 whole-OS *freeze*. A follow-up real-hardware endurance test — run *after* this fix, with the hog already gone — found a **second, distinct failure mode**: an **IMU-process memory leak (~16 MB/min)**, fixed in **#472**. On a 2 GB Pi that leak exhausts RAM → swap-death / OOM, which is the more likely cause of the hard *freeze*. **Revised picture: this PR fixes the comet CPU hog (sustained high CPU, UI starvation, heat); #472 fixes the memory leak (the OOM/freeze). Both are needed for a healthy long observing session.** The original analysis below — including the “not memory” line — predates the endurance finding.

## Problem

During observing — whenever PiFinder has a solve / GPS lock — the comet catalog **continuously pegs a CPU core**. Root cause (py-spy profiled + ablation-proven in a prior session): `calc_comets` advanced all **957 comets** through a per-comet skyfield Kepler-propagation loop (2 `.at(t)` calls each, ~63–70 s total). Worse, while the catalog was still `initialized=False`, the 5 s `TimerMixin` tick kept queueing `do_timed_task` calls on `_task_lock` *during* each multi-second init, so they ran **back-to-back continuously**. Being pure-Python skyfield, it held the GIL and starved the UI loop. Regressed by `4aafd89a` ("replace incremental comet update with full reinit", #391).

Sustained one-core load (heat + GIL-held UI starvation) makes observing sluggish, and was initially diagnosed as the cause of the whole-OS freeze. The endurance follow-up (#472) revised that: the hard freeze was primarily a memory leak in the IMU process. The comet hog is a real, distinct pathology and a compounding factor — but this PR alone does not resolve the OOM/freeze.

## Fix

Replace the per-comet loop with a **single batched numpy/skyfield propagation**. Skyfield 1.45's `propagate()` is array-aware and `mpc._comet_orbits()` builds one `KeplerOrbit` over all rows, so every comet is advanced in one call.

- Positions are read straight from the equatorial state vectors and rotated by `ts.J2000.M` to **reproduce the old `radec(ts.J2000)` framing exactly**.
- The magnitude cut is phrased as `~(mag > 15)` to preserve the old quirk of **keeping unknown/NaN-magnitude comets** (`nan > 15` is False).
- The old per-comet `process_comet` is retained as a tolerant fallback (one bad MPC row can't break the whole batch) and as the test oracle.
- Cadence/threshold logic is unchanged — the per-calc cost was the whole problem; with a fast calc, `initialized` flips immediately and the cadence settles to the intended 293 s.

## Validation

- **Unit benchmark (957 real comets):** `calc_comets` ~63–70 s → **0.13 s (~500×)**, output **byte-identical**: max RA/Dec separation **0.0000″**, |Δmag| = 0, distances match to ~1e-14 AU, identical comet set.
- **Headless A/B** (debug camera + fake GPS, same harness, only the code differs): main-process CPU **~100 % sustained → 1–2 %**; py-spy went from `stumpff`/`propagate`/`_at` dominating to **no comet/skyfield frames**.
- **Real hardware** (service + Tools → Test Mode): AFTER **2 %**, no skyfield frames (the prior session's real-hardware BEFORE was 90 %).

## Tests

`tests/test_comets.py` pins the vectorized path to the per-comet oracle (RA/Dec/mag/distance parity over a spread of eccentricities — also guards against a future skyfield release breaking the batched call) and asserts the NaN-magnitude inclusion quirk and the no-location empty-result path.

`ruff check`/`ruff format` clean; 430 unit tests pass (incl. 4 new); mypy clean on the changed files.

## Related

- **#472** — fixes the IMU-process memory leak (the likely primary cause of the OOM/freeze) plus a contrast-reserve log-spam guard, both found by the endurance test that followed this PR.


brickbots:
Heads-up for reviewers: the follow-up real-hardware endurance test found a **second, distinct failure mode** — an IMU-process memory leak (~16 MB/min) — now believed to be the primary driver of the OOM/freeze this PR was originally credited with fixing. Fixed in #472 (which also rides a small contrast-reserve log-spam guard). I've updated the description above to match. **TL;DR: this PR fixes the comet CPU hog; #472 fixes the memory leak; both are needed for a healthy long observing session.**

mrosseel:
looks good, great catch!

PR #475, opened 2026-06-17
feat(web): scale param + de-duped, hardened screenshot/status API

mrosseel:
## Summary

Improves the web-server screenshot/status API in `python/PiFinder/api_extensions.py` and `python/PiFinder/server.py`:

- **`?scale=N` on `/api/screen`** — optional integer query param (clamped 1..8) that nearest-neighbor upscales the returned PNG, so docs/PR captures of the 128×128 screen stay crisp. `scale=1` (default) is byte-for-byte the old behavior.
- **De-duplicated screenshot serving** — `/image` (server.py) and `/api/screen` both served `shared_state.screen()` as a PNG with slightly different fallbacks. Both now go through a single `get_screen_png()` helper. `/image` keeps its contract (200 `image/png`) so the web-remote live view is unaffected.
- **Hardened `/api/status`** — each `shared_state.*` read is now wrapped individually. A dead/broken multiprocessing manager degrades gracefully: readable fields are still returned, with a clear `error` map of which reads failed, instead of a blanket 500. The happy path is unchanged (no `error` key when everything reads).

## Testing

- `tests/test_api_extensions.py` passes (6/6).
- Added local smoke checks (not committed): scale sizing/clamping (1/4/99→8/0→1), nearest-neighbor pixel preservation, `None`/`BrokenPipeError` screen fallbacks, and a Flask test-client run of `/api/status` against a partly-broken shared state (returns 200 + partial data + `error` field) and `/api/screen?scale=5` (640×640).
- `ruff check` and `ruff format --check` clean on both changed files.

### DIFF
--- a/python/PiFinder/api_extensions.py
+++ b/python/PiFinder/api_extensions.py
@@ -46,6 +46,37 @@ def _png_response(img: Image.Image) -> Response:
     return Response(_pil_to_png_bytes(img), content_type="image/png")
 
 
+def _scale_image(img: Image.Image, scale: int = 1) -> Image.Image:
+    """Nearest-neighbor upscale ``img`` by integer ``scale`` (clamped 1..8).
+
+    scale=1 returns the image unchanged. Nearest-neighbor keeps the pixel grid
+    crisp, which is what docs/PR captures of the 128x128 screen want.
+    """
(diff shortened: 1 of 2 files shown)


jscheidtmann:
Those tests sound useful to have in the codebase

PR #478, opened 2026-06-17
Auto-spawn cedar-detect-server in dev mode

mrosseel:
## Problem

`PFCedarDetectClient` never spawns `cedar-detect-server` — it assumes the
server is already running (as a systemd service on the PiFinder). That holds
on-device, but in a development checkout it means `cedar-detect-server` has to
be started by hand before every run.

## Change

If nothing is listening on the client's port, spawn the bundled
`bin/cedar-detect-server-<arch>` binary as a subprocess (falling back to one on
`PATH`) and tear it down in `__del__`. When a server is already listening — the
systemd service — behaviour is unchanged.

A fresh dev checkout now works without a manual server start, while the
on-device service path is untouched.

### DIFF
--- a/python/PiFinder/solver.py
+++ b/python/PiFinder/solver.py
@@ -16,6 +16,10 @@ import logging
 import sys
 from time import perf_counter as precision_timestamp
 import os
+import platform
+import shutil
+import socket
+import subprocess
 import threading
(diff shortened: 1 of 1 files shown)


brickbots:
Looks good and a nice convenience!  Thank you!

PR #479, opened 2026-06-17
Don't let a hung V4L2 capture freeze the camera process

mrosseel:
## Problem

`CameraInterface.get_image_loop()` calls `self.capture()` on the hot path. A
V4L2 capture can hang indefinitely when the sensor/driver wedges, and because
the call never returns the whole camera process freezes: no new frames, no
solving, and the command-handling loop is never reached — so the camera can't
even be stopped/restarted short of killing the process.

## Change

Run the hot-path capture on a daemon thread with a timeout
(`_capture_with_timeout`). If it doesn't return in time, log and fall back to a
blank frame so the loop keeps running and stays responsive to commands. A
wedged capture now degrades to failed solves instead of a frozen process.

### DIFF
--- a/python/PiFinder/camera_interface.py
+++ b/python/PiFinder/camera_interface.py
@@ -50,6 +50,11 @@ class CameraInterface:
     # Native (camera-driver) auto-exposure, distinct from the solver-driven
     # auto-exposure above. Enabled for daytime alignment via `set_exp:native`.
     _native_ae_enabled = False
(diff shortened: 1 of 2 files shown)


brickbots:
Thanks for spotting and fixing this!  I ran it through code review and Claude spotted that the timeout coupled with the repeated calls could stack these threads.  I don't think that would happen in practice as the overall camera loop should just pause while the thread resolves, but felt like a safe and sensible guard so I added it in.

PR #483, opened 2026-06-19
fix(ui): scroll long status values (IP) instead of truncating

mrosseel:
## What

Status-screen values that are wider than the display (most visibly the IP
address) are currently truncated, so you can't read the whole value. This makes
those long values **scroll horizontally** instead, so the full text is legible.

## Changes

- `python/PiFinder/ui/status.py` — scroll overflowing status values instead of
  truncating them.

Single commit, based directly on `main`.

### DIFF
--- a/python/PiFinder/ui/status.py
+++ b/python/PiFinder/ui/status.py
@@ -10,7 +10,7 @@ import time
 from PiFinder.ui.base import UIModule
 from PiFinder import calc_utils
 from PiFinder import utils
-from PiFinder.ui.ui_utils import TextLayouter, SpaceCalculatorFixed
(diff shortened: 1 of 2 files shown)


jscheidtmann:
Big kudos! 🙏

@brickbots This is a must have for the upcoming release! 😃

mrosseel:
> Big kudos! 🙏
> 
> @brickbots This is a must have for the upcoming release! 😃

if it happens it's super annoying :)

brickbots:
Thanks @mrosseel and hi @jscheidtmann !  I didn't realize there was so much latent need for status screen scrolling 😅 

Very nice work on this, I love that the cost of these QOL fixes is getting low enough that we are knocking a bunch of them out 👍

PR #494, opened 2026-06-25
ci: NixOS release workflow (stable → pifinder-release, beta → dev cache)

mrosseel:
Restores the NixOS release workflow as a focused, trusted workflow (separate from the big NixOS PR #379, and from the testable-PR builder #493).

**What it does** (`workflow_dispatch`, maintainer-triggered, so it runs in brickbots' trusted context with the real `ATTIC_TOKEN` + write token):
- Builds the release closure (`aarch64`, substituting build deps from the `pifinder` dev cache).
- **stable** → pushes the closure to the retained **`pifinder-release`** cache (GC off — devices resolve it months later).
- **beta/prerelease** → pushes to the **`pifinder`** dev cache instead (finite retention, so prereleases stay transient).
- Builds the SD image + migration tarball, tags the source commit, creates the GitHub Release, and writes the `release` entry to the `nixos-manifest` branch.

**Readiness:** the `pifinder-release` cache exists and is public, and the CI token has push (`w:1`) on it — verified server-side. This workflow is inert until the NixOS flake config (`#379`) lands on `main` (it builds `.#nixosConfigurations.pifinder…`).

Note: `softprops/action-gh-release@v1` flags an old-runtime actionlint warning (carried over from the original); worth bumping to `@v2` in a follow-up.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,168 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version (e.g., 2.5.0)'
+        required: true
+      notes:
+        description: 'Release notes'
+        required: true
+      type:
+        description: 'Release type'
+        type: choice
+        options:
+          - stable
+          - beta
+        default: stable
+      source_branch:
+        description: 'Source branch (default: main, use release/X.Y for hotfixes)'
+        required: false
+        default: 'main'
+
+jobs:
+  release:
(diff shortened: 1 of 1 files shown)


mrosseel:
Closing — misplaced. release.yml uses workflow_dispatch, which only fires from the default branch (release), not main, so this would never have been dispatchable here (same lesson as #493/#495). The release workflow (with the beta→dev split) can be re-added to release in its own PR when releases are needed.

PR #501, opened 2026-06-26
Improve focus preview and camera gain control

hjoungjoo:
## Summary

This draft PR is a focused split from my Raspberry Pi 4/CM5 compatibility branch.

It contains only camera focus/gain behavior:

- add a runtime Camera Gain menu next to Camera Exp
- allow set_gain:profile to return to the camera profile default gain
- show the current runtime gain in the gain menu selection state
- avoid applying dark-sky stretch to bright focus frames
- add a raw-frame display fallback for bright/flat focus preview frames

Out of scope for this PR:

- Bookworm install/path changes
- board-specific GPS/UART defaults
- Bluetooth keyboard work
- Korean localization
- MF_PiFinder-specific documentation

## Tested

- python3 -m compileall -q PiFinder
- python3 -m ruff check PiFinder tests
- python3 -m pytest tests/test_main.py tests/test_menu_struct.py -q

### DIFF
--- a/python/PiFinder/camera_interface.py
+++ b/python/PiFinder/camera_interface.py
@@ -65,6 +65,13 @@ class CameraInterface:
         """
         return False
 
+    def get_default_gain(self) -> float:
+        """Return the backend's default startup gain."""
+        profile = getattr(self, "profile", None)
+        if profile is not None and hasattr(profile, "analog_gain"):
+            return float(profile.analog_gain)
+        return float(getattr(self, "gain", 1.0))
+
     def initialize(self) -> None:
         pass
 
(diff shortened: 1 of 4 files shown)


brickbots:
@hjoungjoo Thank you again for a great bit of code here!  We've changed a few things about the preview display, but I like the idea of checking for especially bright frames and displaying the raw image instead.  Would you be able to put up a new PR based on the existing preview.py module, or would you mind if I re-implemented your idea?

Regarding the gain control, is there some situation where it would be expected that the user should adjust this?  The general project philosophy is that less-settings, and more sane defaults, provides a better UX.  Is there specific value for the user in having the gain be adjustable?

hjoungjoo:
> @hjoungjoo Thank you again for a great bit of code here! We've changed a few things about the preview display, but I like the idea of checking for especially bright frames and displaying the raw image instead. Would you be able to put up a new PR based on the existing preview.py module, or would you mind if I re-implemented your idea?
> 
> Regarding the gain control, is there some situation where it would be expected that the user should adjust this? The general project philosophy is that less-settings, and more sane defaults, provides a better UX. Is there specific value for the user in having the gain be adjustable?

I’m currently focusing on other areas, so it’s perfectly fine if you reimplement this part.

The reason I added gain control is that in bright environments or during the daytime, shutter speed alone is not enough. To adjust focus properly in bright conditions, gain adjustment is necessary.

After setting it, it should be returned to the default value again at night.

Also, since the Moon is very bright, gain adjustment can be helpful in that case as well.

PR #508, opened 2026-06-28
Fix state dt tz

alexis-cousein:
I had noticed that when you input time/date "UTC TM" and "LCL TM" were always identical, and that when recovering time from GPS there was the expected gap between them, so I sanity checked some of the datetime timezone usage in the current code.

-The current date/time input routines are self-consistent but they store a datetime in shared_state that contains a timezone reference (to the local timezone the times were specified in), while when using the GPS everything is in UTC) .
-**[bug]** The status screen that displays "UTC TM" just expects the shared state to be in UTC.

Since other consumers of the shared state datetime might make the same mistake, I corrected this at both ends:
-When saving the shared state, always convert to a timezone-aware datetime that is UTC based, so that manual input time and GPS time will always behave similarly.
-Introduced a shared_state query routine utc_datetime in addition to local_datetime to avoid the ambiguous "datetime" (which just returns a valid datetime, but these can be in different timezones); the "TM UTC" visualisation now uses this.

Other bug:
-**[bug]** In set_time use of datetime.now() without explicit timezone to get "today's date" to combine it with the time that was input is just wrong: this yields a naive datetime of the time in the OS timezone; because "today's date in UTC" might be different from the date in the target timezone (which we force before setting it), that may lead us to set the datetime 24 hours off in some cases (which would incorrectly 'prime' the prefilled "date" in the next screen). There, we need to use datetime.now(timezone) to get a day that we *can * combine coherently with the one the user just filled in.

More cosmetic changes:
-replaced pytz.timezone("UTC") by the initialized singleton pytz.utc -- more readable and faster, no need to reconstruct this object time and again.
-gps_fake now explicitly uses a UTC-timezone-stamped datetime.now(pytz.utc) , just to avoid sending "naive" datetimes around.
-catalogs.py was extracting a naive datetime.now() --expressed in the OS timezone-- and just tagged on "UTC" to make a timezone-aware datetime. This is only correct if the OS timezone is UTC, and that is unnecessarily dangerous: datetime.now(pytz.utc) is guaranteed to give you the actual time expressed in UTC, and tagged as such.
-Renamed "GPS LST" ("the LaST time we got a GPS lock") in the status screen to "GPS LKT" (for "GPS LocKTime") to avoid confusion with Local Sidereal Time.


mrosseel:
@alexis-cousein I pushed some follow-ups on top of this branch and opened them as a PR against your branch so you can land them with one click: https://github.com/alexis-cousein/PiFinder/pull/1 — merging it into `fix_state_dt_tz` adds the commits here.

What it adds:
- a regression test locking in the `set_datetime` UTC contract (naive→UTC, aware-local→same instant in UTC, UTC TM ≠ LCL TM at a non-UTC location);
- an ADR + a *Civil time* glossary block documenting the UTC-aware civil-datetime decision, plus docstrings on the accessors;
- `PiFinder/timez.py` as the single sanctioned datetime source, with ruff's `DTZ` (flake8-datetimez) enforced repo-wide and `timez.py` the only exemption — all bare `datetime.now()`/`strptime()`/`fromtimestamp()` routed through it (also fixes a latent relabel bug in `comets.py`).

Verified DTZ-clean + tests passing on this PR's base.

brickbots:
Thank you @alexis-cousein!  This is a great fix.  I added a minor note/question around requirements_dev but this looks ready to merge to me.  Are you feeling like it's good to go?

brickbots:
Sorry, I should have looked a little deeper... What's the time the user is supposed to enter?  If they are entering their local time, we'll need to have their location first.  Otherwise do they have to enter UTC so we can convert to local after getting their location?

If there is a sort of branch here, it would be good to display this to the user:
e.g. "Enter Local Time (UTC)" / "Enter Time (America/Los_Angeles)"?  It's long, but maybe this becomes two lines?

alexis-cousein:
> Sorry, I should have looked a little deeper... What's the time the user is supposed to enter? If they are entering their local time, we'll need to have their location first. Otherwise do they have to enter UTC so we can convert to local after getting their location?
> 
> If there is a sort of branch here, it would be good to display this to the user: e.g. "Enter Local Time (UTC)" / "Enter Time (America/Los_Angeles)"? It's long, but maybe this becomes two lines?

That has nothing much to do with the fix, of course, there are no changes in that area...but you're right we could handle this here to make it slightly better, but it's not trivial.

The current display string says:

```
            _("Enter Local Time"),
```

So I think it's rather self-evident. The intent (and the way I've been using it) is to load the location first and then enter the time.

If the location is unknown, then the routine to set datetime is going to be using the timezone from the default initializer for the Location class:
```
    timezone: Optional[str] = "UTC"
```

This is, in theory, detectable, since if people fill in a location we're going to do:
```
            v.timezone = self.__tz_finder.timezone_at(lat=v.lat, lng=v.lon)
```
which will _never_ return "UTC" (the closest you'll get is "Etc/GMT").

But the trickier parts about supporting this right is not only to change the string to input time. The problem is that in this case the _date_ window will also expect you to enter the UTC date, which may be different from your local date, and doing the conversion is not intuitive at all for a mundane user.

Additionally, the current code expects the date and time input to be "naive" and "local" but will only fetch the timezone to complete it when _**storing**_ the result from the input, not when you actually _started_ the input. Since the shared state may be written by other threads there is even a potential race window if a GPS lock changes the shared state timezone after you started requesting input from the user.

Personally I think it's out of scope fora bug fix PR, we should think about what we want in a different PR.

Choices are:
1. documenting this, and leaving it to the users to decide what to do with it (which is not unreasonable): if the location is not set, "local" time in this screen is UTC time and "local" date is assumed to be the UTC date. You are allowed to set the location later if you want to and the local time will be adjusted accordingly when you do so.
2. adding a warning to the time/date window if the shared state location timezone is still "UTC" -- ugly because the possible race condition is not addressed, unless we can be sure that the GPS stuff isn't going to be changing the location behind our back. We also need to warn that the UTC data may differ from the local date!
3. refuse to allow the user to enter time and date if the shared state location timezone is still "UTC"
4. display timezone "deduced" _just before storing_ and asking user to confirm it's right (or cancel). This is ugly because if we *do* hit the race window or if the timezone is not what the user expected the user is forced to cancel and re-enter time and date.
5. At the *start* of datetime pick up the timezone from the shared state (either already entered through the location or "UTC"). If it's still "UTC" ask the user to confirm he wants to input UTC time **_and date_**, and let the time entry widget's item_definition *also* return a tz_str (in addition to a time_str), so that we know what timezone to use when storing it (we can't use the shared state because it may have changed!)

All of these have consequences for the UI, and I think such refactoring is out of scope for a bug fix PR like this. 

5. really is the only sane one for maximum flexibility, but entails a complete refactoring, and TBH I don't know if it's really worth it, given that for many users far away from UTC deciding what the UTC *date* is is quite the mental gymnastics. 

That may make 3. (refusing manual time entry if the location was not yet set) more appealing, especially if we expect that users would always choose to cancel and enter the location *rather* than do the mental gymnastics to entre UTC time and date. 

While we mull over this, I think 1. makes the most sense.

All that being said, there is no harm in changing "Local" time to display the *actual* timezone instead. That's something that addresses some of the issue, but not all of it (since it's not only the *time* but the date that can be influenced by the timezone, and since there's still a race condition unless you go with solution #5.)

alexis-cousein:
I feel like it's good to go, yes.

brickbots:
Thanks @alexis-cousein for the contribution and the thoughtful note about some future options!  I'm going to merge this and work on putting a guard in place to make sure location is set before time (your option 3) :+1:

PR #510, opened 2026-06-29
Import third-party CSV observing lists (lenient headers + coordinate formats)

mrosseel:
## Problem

CSV is a supported observing-list format, but `read_csv` only accepts the exact
columns and sexagesimal coordinate strings that `write_csv` emits. A list
exported from another tool — lowercase headers, decimal-degree coordinates —
silently loads as nameless objects at RA 0 / Dec 0 instead of importing.

A real example that fails today:

```
name,ra,dec,mag
mess003,205.8583333,28.2441667,6.3
mess013,250.6666667,36.4111111,5.8
```

## Change

CSV exists to import lists made elsewhere, so `read_csv` is now lenient:

- **Headers** are matched case-insensitively through an alias table
  (`name`, `ra`/`ra_deg`, `dec`/`de`, `mag`/`magnitude`/`vmag`, `type`/`obj_type`,
  `catalog`, `seq`, …). Column order is free; values are stripped.
- **Coordinates** parse as decimal degrees, `HH:MM:SS` / `±DD:MM:SS`, or the
  existing `Xh Xm Xs` / `±D° M' S` — first hit wins. A bare decimal RA is degrees
  (0–360) by default; an hours-named header (`RA_h`, `RA_hours`) reads it as hours.
- **Detection** recognizes a CSV header row (`Name` plus `RA`/`Dec`, any case),
  so a mis-named `.txt` export is still read as CSV.

Resolution: a row whose name matches a catalog designation resolves to that
catalog object (images, descriptions, logs); otherwise the row's own coordinates
become an OBS target. Name matching now reuses the shared `ui_utils.normalize`
helper, so spacing/case variants (`M 3`, `M3`, `NGC 224`) resolve consistently
regardless of how the catalog stores the designation — previously `M 3` missed
because the catalog stores Messier as `M13`/`M3` (no space). To keep your own
coordinates for an otherwise-matchable name, use a non-designation label (e.g.
`_M 3`); `normalize` keeps the underscore.

CSV is documented as an import format.

## Tests

`tests/test_obslist_formats.py` adds decimal-degree/lowercase-header,
header-alias, and colon-coordinate import cases plus two detection cases; the
sexagesimal round-trip is preserved. `test_obslist_formats.py` and
`test_obslist_resolve.py` pass (47 tests); mypy is clean on the changed module.

## Docs

- `docs/source/user_guide.rst`: new "Importing a CSV list" subsection (columns,
  the three coordinate forms, resolution + opt-out).
- `docs/ax/catalog/obslist-formats/README.md`: CSV marked **read (import)**;
  added a CSV field reference.

https://claude.ai/code/session_01Xc4W7jF784MCTQ72PetHcp


jscheidtmann:
How about RA in hour angle? The import process should ask, if all RA are less than 24 or provide an option on the upload page.

mrosseel:
> How about RA in hour angle? The import process should ask, if all RA are less than 24 or provide an option on the upload page.

the import process can't ask as its automatic: if there's a csv file on the sd card it's shown in the ui after automatic importing, no chance to ask the user something. You did give me an idea though, the column names can give us the hint that it should be read in hours. I'll add that in.

brickbots:
Hi @mrosseel Thanks for making this even better, and for the first improvement for 2.6.1 🎉   Are you feeling that this is ready to merge?

mrosseel:
@brickbots good to merge

PR #511, opened 2026-06-30
refactor(catalog): single-source the object-type code set from OBJ_TYPES

mrosseel:
## What

`OBJ_TYPES` (`python/PiFinder/obj_types.py`) becomes the single source of truth for the object-type code set, which was hand-duplicated in the filter menu and the docs (and had drifted).

- **Filter menu** (`ui/menu_structure.py`) — the Type filter items are now generated from `OBJ_TYPES` (drops the hardcoded list of code/label pairs).
- **`obj_types.py`** — reordered to the filter-menu display order (iteration order only; lookups are unaffected) and documented as the canonical source.
- **Drift guard** (`tests/test_obj_types_docs.py`) — fails if anything drifts out of sync with `OBJ_TYPES`:
  - `default_config.json`'s `filter.object_types` (the default Type-filter selection) must equal the `OBJ_TYPES` set, so a newly added type can't be silently off by default;
  - the docs "Object type codes" table must match `OBJ_TYPES` (no-ops where that table isn't present — it ships in the obslist docs PR — and activates once both land on `main`).

## User-facing change

The Type filter menu adopts the same labels already used on the object-detail screen (they previously disagreed):

| Before | After |
|---|---|
| Cluster/Neb | Cluster + Neb |
| P. Nebula | Planetary |
| Double Str | Double star |
| Triple Str | Triple star |
| Unknown | Unkn |

These msgids already exist and are translated in every locale (they are the detail-screen labels).

## Test plan

- 221 menu/UI tests, catalog/filter/config unit tests, and smoke tests pass; ruff clean.
- New drift-guard test.

### DIFF
--- a/python/PiFinder/obj_types.py
+++ b/python/PiFinder/obj_types.py
@@ -3,22 +3,25 @@ def _(key: str) -> str:
     return key
 
 
+# The single source of truth for the object-type code set and its labels.
+# Iteration order is the order shown in the Type filter menu (which is generated
+# from this dict); lookups (e.g. the object-detail screen) are order-independent.
 OBJ_TYPES = {
     "Gx": _("Galaxy"),  # TRANSLATORS: Object type
     "OC": _("Open Cluster"),  # TRANSLATORS: Object type
+    "C+N": _("Cluster + Neb"),  # TRANSLATORS: Object type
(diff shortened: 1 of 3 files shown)


brickbots:
Thanks @mrosseel great bit of cleanup here!

PR #516, opened 2026-07-03
ci: populate manifest on every release build + record migration tarball URL

mrosseel:
Follow-up to #514/#515. Makes the Release workflow feed the update manifest, and adds robustness.

**1. Manifest is populated on every build, including fork-source builds.** The "Update generated manifest branch" step was `same-repo`-guarded, so `source_repository`-driven builds (the current way we build `mrosseel:nixos`) published assets but never recorded a manifest entry — so `software.py`'s channel UI never saw the release. Now the release job checks out the base repo (for the trusted manifest scripts + correct push `origin`), takes the built source SHA from the build job, and always runs the manifest step.

**2. Manifest release entries now carry the migration tarball.** Added `migration_url` + `migration_sha256_url` to `update_manifest.py` release entries (pointing at this repo's release assets). This lets the **migration** path select a release from the manifest (stable→beta→unstable) instead of a hardcoded URL — companion app-side change to follow.

**3. Fix image-artifact corruption when both builders run.** If the native build is slow, `native-wait` times out and the hosted fallback also runs; both upload an identically-named `.img.zst`. The old `download-artifact … merge-multiple` merged them → `zstd` "Data corruption detected" (this failed the v3.0.0-beta run). Now native/hosted images download to separate dirs and the assemble step prefers native.

Validated with actionlint (only the known self-hosted-label warning) and py_compile.

### DIFF
--- a/.github/scripts/update_manifest.py
+++ b/.github/scripts/update_manifest.py
@@ -61,7 +61,11 @@ def set_available(entry: dict) -> dict:
 
 
 def replace_entry(entries: list[dict], predicate, entry: dict) -> list[dict]:
-    return [item for item in entries if not predicate(item)] + [entry]
+    # Newest first: consumers (release channels in the update UI, and the
+    # migration target lookup) take the head of the list as the current entry.
+    # The unstable channel is re-sorted after this, so prepending is neutral
+    # there.
+    return [entry] + [item for item in entries if not predicate(item)]
 
 
 def sort_unstable(entries: list[dict]) -> list[dict]:
@@ -140,6 +144,8 @@ def update_release(args: argparse.Namespace) -> None:
         "source_sha": args.sha,
         "version": args.version,
         "store_path": args.store_path or None,
+        "migration_url": args.migration_url or None,
+        "migration_sha256_url": args.migration_sha256_url or None,
     }
     set_available(entry)
     manifest["channels"][channel] = replace_entry(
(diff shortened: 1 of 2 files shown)


mrosseel:
Two fixes pushed after self-review:
1. **Tag now points at the built source commit** (`git tag $TAG $SOURCE_SHA` after fetching it). The base-repo checkout is at the workflow's ref, so the previous `git tag $TAG` on HEAD would have tagged the release branch instead of the built `source_branch`.
2. **Hosted-only fallback no longer crashes the assemble step**: `find img-native img-hosted` failed under `set -euo pipefail` when only one artifact dir existed. Both dirs are now pre-created and the lookup uses `find -print -quit` with an explicit empty-result error.

mrosseel:
Two more fixes from review:
1. **`bash .github/scripts/publish_manifest.sh`** — the script is checked in as mode 100644 (non-executable), so the bare invocation would fail with Permission denied; the PR-build workflow already uses the `bash` prefix for the same reason.
2. **`replace_entry` now prepends** so release channels are ordered **newest-first**. Previously a second stable release would append (`[old, new]`) and manifest consumers taking the head — the update UI's channel lists and the migration target lookup — would keep serving the older release. The unstable channel is re-sorted after insertion, so this is neutral there. Verified with a double-release simulation: `stable == [v3.1.0, v3.0.0]`, re-releases of the same tag dedupe in place.

PR #517, opened 2026-07-03
feat(migration): source the migration tarball from the update manifest (stable→beta→unstable)

mrosseel:
Replaces the hardcoded migration tarball URL with a lookup in the update manifest, so migration follows releases automatically instead of a pinned `v3.0.0-migration` URL.

## What changed
- New `_migration_version_info_from_manifest()`: fetches `update-manifest.json` (brickbots `nixos-manifest`), walks channels **stable → beta → unstable**, and returns `version_info` (`version`, `type`, `migration_url`, `migration_sha256_url`) for the **first available entry that carries a migration tarball**. Skips unavailable entries and entries without a tarball.
- **Unlock (7× square):** now offers the first-available manifest target (was the hardcoded `_MIGRATION_VERSION_INFO`); shows "No release found" if the manifest has none.
- **Public/automatic path:** still gated by `nixos_for_everyone` (unchanged rollout control), but the tarball now comes from the manifest rather than the gate's `nixos_url`. `_fetch_migration_config()` no longer requires `nixos_url` — it only carries the gate flag.
- Removed the hardcoded `_MIGRATION_VERSION_INFO`.

Depends on the manifest carrying `migration_url`/`migration_sha256_url`, added by the release workflow in #516.

## Tests
- 37 pass. Added `TestFetchUpdateManifest` + `TestMigrationVersionInfoFromManifest` (channel priority, beta/unstable fallback, skip unavailable, skip entries without a tarball, network/JSON failure → None). Updated the two obsolete `_fetch_migration_config` url-requirement tests.

Note: the new "No release found" string needs `nox -s babel` extraction before translation.

### DIFF
--- a/python/PiFinder/ui/software.py
+++ b/python/PiFinder/ui/software.py
@@ -28,25 +28,25 @@ REQUEST_TIMEOUT = 10
 MIGRATION_GATE_URL = (
     "https://raw.githubusercontent.com/brickbots/PiFinder/release/migration_gate.json"
 )
+UPDATE_MANIFEST_URL = (
+    "https://raw.githubusercontent.com/brickbots/PiFinder/"
+    "nixos-manifest/update-manifest.json"
+)
 
 # Secret unlock: 7x square button
 _UNLOCK_SEQUENCE = ["square"] * 7
 
-_MIGRATION_VERSION_INFO = {
-    "version": "3.0.0",
-    "type": "upgrade",
-    "migration_size_mb": 292,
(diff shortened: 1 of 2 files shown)


mrosseel:
Added `migration_size_mb` to the manifest-selected version_info via a HEAD request on the tarball (follows GitHub's 302 to read the real Content-Length; omitted on failure so the confirm screen shows "?"). Previously the confirm screen would always have shown "Download: ?MB". Tests extended (39 pass).

PR #518, opened 2026-07-03
docs: fold field feedback into the Polar Alignment guide

brickbots:
## Summary

Early users of the Polar Alignment tool reported back with the technique that actually works in the field, plus the failure mode that most often ruins a session. This PR folds that feedback into the user guide's Polar Alignment section.

- **New "Getting a good measurement" sub-section** with per-mount advice:
  - *EQ mount*: point the Dec axis ~7–10° off the mount's polar axis (wherever it points now), sweep 30–45° in RA with the middle capture screen-vertical and the ends 15–22° either side, and use **Roll Off**.
  - *Platform*: aim within ~5° of Polaris, keep **Roll On**, and set the screen vertical at mid-travel so the alt/az arrows match real directions.
  - Realistic accuracy expectations: 20–30 arcmin typical, down to ~10 with a very rigid PiFinder-to-scope connection — not a polar-scope replacement, but plenty when the pole is hidden or to get Polaris into a polar scope's reticle.
- **New note at the adjustment stage**: lock *both* mount axes before touching the adjusters — a slipping RA clutch (unlocked moments earlier to rotate between captures) was the most common field failure — and on a platform, adjust only the platform, never the scope.
- **Prerequisites**: turn off sidereal tracking before starting.
- **Reversed one piece of existing advice**: step 1 used to say to aim *well away* from the pole. Field experience contradicts this — flexure (felt mostly in roll) grows with how much the PiFinder's attitude changes between captures, so moderate near-pole sweeps beat wide ones. Step 1 now points at the new tips sub-section.

## Not included (possible follow-up)

The same discussion suggested putting a short version of the lock-your-clutches / tracking-off reminder **on the ADJUST screen itself**, varying by EQ/Alt-Az mode. That's a code change to `ui/polar_align.py`, left out of this docs-only PR.

## Testing

- `sphinx-build -b html -n` (nitpicky mode) runs clean — no warnings, both new `:ref:` targets resolve.

### DIFF
--- a/docs/source/user_guide.rst
+++ b/docs/source/user_guide.rst
@@ -977,7 +977,8 @@ It works by solving the sky at two or three points while you rotate around that 
 them — turning the platform, or slewing the mount in right ascension only — then working out
 where the axis points from how the view shifts.  The measurement is only as good as the
 solves behind it, so before you start make sure the PiFinder has a GPS lock, is focused, and
-is solving reliably.
+is solving reliably.  Turn off your mount's or platform's sidereal tracking as well — the
+measurement assumes the scope holds still except when you rotate it yourself.
 
 .. image:: images/user_guide/polar_align_intro_docs.png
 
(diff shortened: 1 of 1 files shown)


brickbots:
@mrosseel and @alexis-cousein Should I merge this and move it to the release branch so the default docs get updated?  It seems like this might be useful even if it's still under development.

PR #520, opened 2026-07-04
fix(migration): generate NM wifi keyfiles in Python pre-reboot with correct SSID encoding

mrosseel:
Fixes the bug that left every migrated device unable to join its wifi network: the busybox-shell converter in `nixos_migration_init.sh` wrote SSIDs into NetworkManager keyfiles as **hex** byte lists (`ssid=61;70;6f;…`), but NM's keyfile format only parses **decimal** byte lists — the garbage was stored as a literal-string SSID, so the device scanned for a network literally named `61;70;6f;6c;6c;6f;` ("apollo"). Diagnosed on real hardware (profiles repaired in place with `nmcli`); a fun corollary: an SSID whose hex pairs are all decimal digits parsed "successfully" into different garbage (`"0x20"` → phantom `N ` network).

## What changed
- Keyfiles are now generated **in Python on Debian, before reboot** (`PiFinder/nixos_migration_wifi.py`), instead of in busybox shell after the rootfs is staged — unit-testable, and the initramfs just copies the pre-staged files.
- `ssid=` emits **decimal** byte lists (`97;112;…`), the documented format.
- wpa_supplicant's **unquoted-hex SSID form is decoded** (it's how SSIDs with special characters are stored) — previously double-mangled. Quoted hex-lookalike names stay verbatim; non-UTF-8 SSID bytes round-trip via surrogateescape.
- Pre-existing NM keyfiles on the Debian side are still preserved as-is.

46 unit tests (`test_nixos_migration_wifi.py`): parsing (quoted/unquoted/hex PSK/comments/multiple networks), encoding, filename sanitisation, keyfile structure, end-to-end hex-SSID conversion.

This is pre-migration (Raspbian-side) code: `sys_utils.start_nixos_migration` runs `python/scripts/nixos_migration.sh` from the device's checkout, which now pre-stages the keyfiles and builds them into the migration initramfs.


mrosseel:
Superseded by #521 — same fix, minimal form (one file, ~30 lines, keeps the existing shell converter and corrects only the encoding). The Python pre-reboot generator in this PR remains on the fork's `nixos` branch and rides the eventual upstreaming instead.

PR #521, opened 2026-07-04
fix(migration): generate NM wifi keyfiles in Python pre-reboot with correct SSID encoding

mrosseel:
Fixes the bug that leaves every migrated device unable to join its wifi network, by replacing the busybox-shell keyfile converter with a unit-tested Python generator that runs on Debian **before** reboot; the initramfs just copies the pre-staged keyfiles.

## The bug
The shell converter wrote SSIDs into NetworkManager keyfiles as **hex** byte lists (`ssid=61;70;6f;…`), but NM's keyfile format only parses **decimal** — the garbage was stored as a literal-string SSID, so the device scanned for a network literally named `61;70;6f;6c;6c;6f;` ("apollo"). Diagnosed on real hardware (profiles repaired in place with `nmcli`). Corollary: an SSID whose hex pairs were all decimal digits (`"0x20"`) parsed "successfully" into different garbage — a phantom `N ` network.

## The fix
- `PiFinder/nixos_migration_wifi.py`: parses wpa_supplicant.conf and emits NM keyfiles with **decimal** byte lists; also decodes wpa_supplicant's unquoted-hex SSID form (how SSIDs with special characters are stored) with non-UTF-8 bytes round-tripping via surrogateescape; hex PSKs kept verbatim; filenames sanitised.
- `nixos_migration.sh` pre-stages the keyfiles into the initramfs; `nixos_migration_init.sh` drops its shell converter and just copies them (pre-existing NM keyfiles on the Debian side still preserved as-is).
- 46 unit tests covering parsing, encoding, sanitisation, keyfile structure, and end-to-end hex-SSID conversion.

Generating on full Debian instead of in busybox after the rootfs is staged is what makes this testable — the class of bug being fixed is exactly the kind that untested initramfs shell accumulates.


mrosseel:
Added a second fix to the bundle: `nixos-pr-build.yml` still pinned the **rolled-back** Attic cache key (`Vkem…`, from the aborted S3-cutover rotation) — the cache signs with the original `8UU…`, so the hosted-fallback job in testable-PR builds could verify nothing from `cache.pifinder.eu` and silently rebuilt everything from scratch. Since `pull_request_target` executes this workflow from `main`'s copy, main's key is the live one for all PR builds. One line: `Vkem` → `8UU`.

mrosseel:
Third fix, from a live 2GB-board migration attempt: the initramfs mounts /tmp with the tmpfs default cap (50% of RAM), so copying a 1GB tarball to RAM failed on a 2GB board even though total RAM sufficed — and the failure stranded the device in a console-only debug shell. Now: (1) tmpfs cap raised to 90% (the explicit MemAvailable checks remain the guard); (2) a pre-reboot feasibility gate in nixos_migration.sh fails on the running OS, never in the initramfs; (3) any initramfs failure **before formatting** restores the old OS's boot config and reboots back instead of dropping to a shell.

mrosseel:
Scope restored: this PR is the wifi keyfile fix (+ the Attic key one-liner). The RAM/stranding fixes moved to their own PR: #523.

PR #523, opened 2026-07-05
fix(migration): guard RAM failures and install boot-on-ext4 tarballs safely

mrosseel:
Follow-up migration hardening from the first live failures. This PR is stacked on #521; keep both PRs open and merge #521 first.

- Raise the initramfs tmpfs cap to 90%; explicit memory checks remain the safety gate.
- Reject tarballs that cannot fit in board RAM before changing boot configuration.
- Restore the previous OS boot configuration automatically after any pre-format failure.
- Treat formatting as the explicit destructive boundary; later failures enter the debug shell.
- Handle the actual migration tarball layout correctly: top-level boot is firmware, while rootfs/boot contains extlinux, kernels, and initrds on ext4.
- Stage firmware aside before moving rootfs so the two boot directories cannot collide.
- Verify config.txt on FAT and extlinux.conf on ext4.
- Convert unexpected set -e exits into a visible migration failure instead of allowing init to exit and panic the kernel.

Related release protection: #522 enforces the migration tarball RAM budget.

Verification: the nox check passes on this PR.


mrosseel:
Round-two field failure, second commit: the init script predates the boot-on-ext4 layout. The tarball's top-level `boot/` (firmware payload) and `rootfs/boot` (extlinux + kernels, now non-empty) collided on the same path — and the failing `mv` had no `|| fail`, so `set -e` killed init: **kernel panic** instead of the debug shell. Fixes: firmware payload staged aside before the rootfs move; both partition verifications now check what each actually needs (config.txt on FAT, extlinux.conf on ext4 — two places demanded extlinux on FAT, where it no longer lives); an EXIT trap routes any uncaught failure into fail() (disarmed before the deliberate final reboot), so a panic is structurally impossible.

mrosseel:
**Restacked on top of #521** (this PR now includes #521's commits — merge #521 first, then this fast-forwards). Reason, from the live rehearsal: both PRs carried full copies of `nixos_migration_init.sh` with divergent regions, and a multi-branch merge resolved one block from the stale side — the migration failed at the final (already-fixed-here) verification. Single lineage makes that structurally impossible, and also removes this branch's inherited copy of the old hex SSID converter. Second commit here additionally fixes the initramfs for the boot-on-ext4 layout (firmware payload staged aside; per-partition verification; EXIT trap so failures can never kernel-panic).

PR #524, opened 2026-07-05
fix(gps): correct UBX NAV-SVINFO and NAV-SAT satellite decoding

mrosseel:
## Problem

The satellite counts shown by PiFinder (e.g. `GPS 20/0` on the status screen) were wrong on all GPS receiver generations, in two different ways:

- **protVer < 15 receivers** (NEO-6/7, early M8 — NAV-SOL + NAV-SVINFO from gpsd): `_parse_nav_svinfo` read every per-satellite field one byte off. The UBX repeated block is `chn, svid, flags, quality, cno, elev, azim`, but parsing started at `chn`. So satellite IDs were channel numbers, the per-sat used flag was `svid & 1` (random), and "C/N0" was really the 0–7 quality indicator. Since quality is ≥1 even for idle SBAS/QZSS channels still searching, the *seen* count was inflated to near the full channel count (the mysterious "20 seen" with nothing tracked), and `uSat` was garbage — which is presumably why it was treated as stale and ignored in favour of NAV-SOL.
- **protVer ≥ 15 receivers** (later M8, M9, M10 — NAV-PVT + NAV-SAT; M10 has no NAV-SOL/SVINFO at all): the used count stayed **0 forever** despite a valid fix, because NAV-PVT's `numSV` was parsed but never surfaced, and NAV-SOL never arrives on these units.

## Fix

- `_parse_nav_svinfo`: shift field offsets by one per the u-blox spec; decode elev/azim as signed; count only satellites with an actual signal as seen. With correct offsets the `svUsed` flag is valid, so SVINFO now reports a trustworthy `uSat`.
- `gps_ubx.py`: report `uSat` from SVINFO instead of the stale-info workaround; update the used count from NAV-PVT `numSV`; actually set `got_sat_update` when NAV-SAT arrives so the SVINFO fallback defers to it as intended.
- `_parse_nav_sat`: take used from the dedicated `svUsed` bit (bit 3) instead of the quality heuristic (which also counts tracked-but-unused sats); don't count zero-signal entries as seen; decode elev/azim as signed.

## Verification

- New unit tests (`tests/test_gps_ubx_parser.py`) with spec-built payloads for both message types, including the idle-channel inflation case.
- Live-verified against a real MAX-M8 (PROTVER 14) via gpsd: the patched parser's SVINFO output (`nSat 11 / uSat 6`, PRNs, C/N0) exactly matches NAV-SOL `numSV` and gpsd's own SKY decode; the old code produced the inflated counts on the same stream.
- Full unit suite, ruff, and mypy pass.

The NAV-PVT/NAV-SAT path is verified against the spec and unit tests only — a sanity check on a unit with a newer (M9/M10) module would be welcome.

https://claude.ai/code/session_01NrdGJ1s3jmA9qzV8HAsdZs

### DIFF
--- a/python/PiFinder/gps_ubx.py
+++ b/python/PiFinder/gps_ubx.py
@@ -32,14 +32,18 @@ async def process_messages(
         elif msg_class == "NAV-SVINFO" and not got_sat_update:
             # Fallback satellite info if NAV-SAT not available
             if "nSat" in msg:
-                # uSat is also in the message but contains stale info
                 sats_seen = msg["nSat"]
+                sats_used = msg["uSat"]
                 sats[0] = sats_seen
+                sats[1] = sats_used
                 gps_queue.put(("satellites", tuple(sats)))
-                logger.debug("Number of sats (SVINFO) seen: %i", sats_seen)
+                logger.debug(
+                    "Number of sats (SVINFO) seen: %i, used: %i", sats_seen, sats_used
+                )
 
         elif msg_class == "NAV-SAT":
             # Preferred satellite info source - not seen in the current pifinder gps versions
+            got_sat_update = True
             sats_seen = msg["nSat"]
             sats_used = sum(
                 1 for sat in msg.get("satellites", []) if sat.get("used", False)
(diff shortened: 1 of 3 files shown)


brickbots:
@mrosseel is this still waiting on some additional commits, or is it ready to merge?

PR #526, opened 2026-07-10
perf(catalogs): skip re-filtering unchanged catalogs on list open

mrosseel:
## What & why

Opening any object list (Objects › By Catalog › NGC, etc.) is slow — measured **~0.58 s per open** on a Pi. Profiled on-device with cProfile: **604,680 `apply_filter` calls across 4 opens** of the NGC list.

**Root cause:** `UIObjectList` → `refresh_object_list()` calls `Catalogs.filter_catalogs()`, which re-applies the filter across **every** catalog — all ~151k objects — even when you only want to view a single catalog like NGC (7.8k). The existing per-object result cache avoids *recomputation*, but not the O(151k) *iteration* itself, so you pay it on every open.

## The fix

Each `Catalog` already records `last_filtered`, and the filter records `dirty_time` (advanced only when a filter parameter actually changes) — but the two were never compared. Guard `filter_objects()` to return the cached filtered list when the catalog is already filtered against the current criteria:

- `filter_catalogs()` goes from **O(objects)** to **O(catalogs)**.
- Repeat opens with an unchanged filter drop from **~0.58 s to microseconds**.
- A changed filter param (bumps `dirty_time`) still triggers a full re-filter — results unchanged.

### Correctness: invalidate on object mutation

The cache key is really *(filter params, object set)*. `dirty_time` covers the params; the object set can change without it (deferred background load, comet refresh, batch build — all via `add_object`/`add_objects`). Those now reset `last_filtered`, so a catalog that gained objects after its first (empty) filter can't keep serving a stale list. `last_filtered` moves to `CatalogBase` as the single source of truth.

## Files
- `PiFinder/catalogs.py` — dirty guard in `Catalog.filter_objects()`; drop duplicate init.
- `PiFinder/catalog_base.py` — own `last_filtered`; reset it in `add_object`/`add_objects`.

## Validation
- `py_compile` + `ruff check` clean.
- Behavior verified by reasoning through: unchanged filter → cache hit; changed param → `dirty_time` advance → re-filter; deferred/comet/batch add → `last_filtered` reset → re-filter.
- Full test suite via CI.

## Follow-up (not here)
The NEAREST-sort branch re-applies the filter to the already-filtered menu items; skippable when the list is already filtered, but load-bearing for the unfiltered recent/custom paths — left alone to keep this PR focused.

### DIFF
--- a/python/PiFinder/catalog_base.py
+++ b/python/PiFinder/catalog_base.py
@@ -85,6 +85,11 @@ class CatalogBase:
         self.__objects: List = []
         self.id_to_pos: Dict[int, int] = {}
         self.sequence_to_pos: Dict[int, int] = {}
+        # Wall-clock time this catalog's objects were last filtered; compared
+        # against the filter's dirty_time to reuse the cached filtered list.
+        # Reset to 0 on any object mutation so a stale cache can't survive a
+        # deferred load, comet refresh, or batch add (see Catalog.filter_objects).
+        self.last_filtered: float = 0
 
     def get_objects(self) -> ROArrayWrapper:
         return ROArrayWrapper(self.__objects)
@@ -98,6 +103,7 @@ class CatalogBase:
         self._update_id_to_pos()
         self._update_sequence_to_pos()
         assert self.check_sequences()
+        self.last_filtered = 0  # objects changed -> invalidate filter cache
 
     def _add_object(self, obj):
         self.__objects.append(obj)
@@ -112,6 +118,7 @@ class CatalogBase:
         self._update_id_to_pos()
         self._update_sequence_to_pos()
         assert self.check_sequences()
+        self.last_filtered = 0  # objects changed -> invalidate filter cache
 
     def _sort_objects(self):
         self.__objects.sort(key=self.sort)
--- a/python/PiFinder/catalog_base.py
+++ b/python/PiFinder/catalog_base.py
@@ -120,6 +120,20 @@ class CatalogBase:
         assert self.check_sequences()
         self.last_filtered = 0  # objects changed -> invalidate filter cache
 
+    def clear_objects(self):
+        """
+        Remove all objects and reset the sequence/id indexes.
+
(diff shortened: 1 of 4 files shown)


brickbots:
Pushed `7a63f5e6` with a fix for a cache-invalidation gap found in review, plus tests.

**The gap:** the guard's cache key is *(filter params, object set)*, but only object *additions* reset `last_filtered`. `CometCatalog.refresh()`/`init_comets()` clear the object list directly, so after a comet refresh the CM catalog kept serving its stale pre-clear filtered list — indefinitely if the recalc failed (e.g. no internet at the telescope), showing ghost comets with wiped sequence indexes. Verified empirically: same probe returns `[]` on `main`, stale `[1, 2, 3]` on this branch pre-fix.

**The fix:**
- `CatalogBase.clear_objects()` — single clear path that resets indexes *and* invalidates the cache; comet refresh/init now use it.
- Empty-catalog check moved above the cache guard in `Catalog.filter_objects()`, so an emptied catalog always yields `[]` even if a future clear path forgets to invalidate. Both checks are O(1), perf win intact.
- `tests/test_catalog_filter_cache.py` pins the cache contract (reuse on unchanged filter; invalidation on param change / add / clear; mutation + `mark_dirty`). The two clear-related tests fail without the fix, pass with it. Full smoke+unit suite: 713 passed.

Also noted in review, deliberately *not* in this PR (staleness pre-exists in the per-object cache, unchanged by this PR): logging an object doesn't re-filter observed-status lists, and altitude verdicts stay frozen between filter edits. Follow-up planned.

brickbots:
@mrosseel This is a great UX improvement!  Thank you!  Are you okay to merge this with the extra commit from the Claude review?  Would you rather we revert that comet update and merge as you left it?

mrosseel:
No good catch, didn't list it as mergeable because Neede to test. Merging is ok if you tested, and maybe do an extra check if all places for through the methods for clear etc

PR #529, opened 2026-07-12
Fix AttributeError when importing eyepieces from DeepskyLog

brickbots:
## Summary

Importing equipment from DeepskyLog on the web interface crashes with an "Internal Server Error" as soon as the import contains a new eyepiece. This is a v2.5.1 → v2.6.0 regression introduced by the Flask migration (#331), which changed the eyepiece-add call from

```python
cfg.equipment.add_eyepiece(new_eyepiece)
```

to

```python
cfg.equipment.eyepieces.add_eyepiece(new_eyepiece)
```

`Equipment.eyepieces` is a plain `List[Eyepiece]` with no `add_eyepiece` method, so the first new eyepiece raises `AttributeError`. Since `cfg.save_equipment()` comes after the import loop, nothing gets saved — telescopes already collected in the same request are lost too.

This PR restores the pre-migration call on `Equipment` itself (`equipment.py:87`), which appends the eyepiece **and** keeps the list sorted by focal length. The manual add-eyepiece route was unaffected (it already calls the correct method at `server.py:736`).

## Regression tests

`tests/test_server_dsl_import.py` drives the real `/equipment/import_from_deepskylog` route through Flask's test client, with `pydeepskylog` and `config.Config` mocked out and a **real** `Equipment` instance so the plain-list bug reproduces:

- `test_import_adds_new_eyepieces` — new eyepieces are added (sorted by focal length, HTML entities decoded) and the config is saved. With the bad call reintroduced, this fails with `AttributeError: 'list' object has no attribute 'add_eyepiece'` — verified.
- `test_import_skips_existing_eyepiece` — an eyepiece already in the config is not duplicated.

## Test plan

- New regression tests pass (and fail on the un-fixed code)
- `pytest -m "unit or smoke"` passes (785 passed)
- `ruff check` / `ruff format --check` pass

### DIFF
--- a/python/PiFinder/server.py
+++ b/python/PiFinder/server.py
@@ -680,7 +680,7 @@ class Server:
                     try:
                         cfg.equipment.eyepieces.index(new_eyepiece)
                     except ValueError:
-                        cfg.equipment.eyepieces.add_eyepiece(new_eyepiece)
+                        cfg.equipment.add_eyepiece(new_eyepiece)
 
                 cfg.save_equipment()
                 self.ui_queue.put("reload_config")
--- a/python/tests/test_server_dsl_import.py
+++ b/python/tests/test_server_dsl_import.py
@@ -1,0 +1,99 @@
+"""Regression tests for the DeepskyLog equipment import route.
+
+The Flask migration (#331) changed the eyepiece-add call in
+``/equipment/import_from_deepskylog`` from ``cfg.equipment.add_eyepiece(...)``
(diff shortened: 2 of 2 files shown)


mrosseel:
Seems like this cold use a unit test/integration test. Are these run regularly btw?

brickbots:
Yes indeed!  I'll make sure to add a test.  Web interface tests don't run as part of the CI/CD, so it's something we'll need to do more of manually as part of a release :+1:

PR #544, opened 2026-07-19
SQM: scale-aware photometry, tracked black level, wizard capture fixes

mrosseel:
Three stacked fixes to the SQM pipeline, found while cross-calibrating two PiFinders (mrosseel's imx462/HQ units and rich's imx296 + imx462) against hand-held reference meters over 19 referenced sweeps.

## 1. Scale-aware photometry & wing geometry (`fix(sqm): scale photometry...`)

All photometry radii (aperture 5, annuli 10–18, wing-estimator plateau/sky ring) are tuned for the ~1.0-scale Bayer-green photometry images (imx462: 490px vs the 512px solve image). The mono imx296 does photometry on its full-res 1088px frame (scale 2.125), where that geometry sits *inside* the star profile: measured growth curves show the r=5 aperture holds only 83–94% of a star's flux, background annuli land on the PSF, and the wing estimator's own sky ring eats the wings — it measures f≈1 and stays permanently inert.

Fix: convert all radii from solve-image pixels to the photometry image's pitch, and give `WingEstimator` a `set_scale()`. imx462/imx290 round to essentially unchanged values; the imx296's median error vs the reference meter goes **−0.20 → 0.00** on the reference sweeps.

## 2. Tracked black level supersedes static bias (`fix(sqm): tracked black level...`)

The imx296's OB clamp delivered a black level of ≈**56 ADU** during the reference sweeps (background-vs-exposure intercepts 55.4–56.9; near-dark 25ms frames floor at 56) while the profile says 60 and the device's same-night wizard calibration says 58. At dark-sky signal levels a 2–4 ADU over-subtraction produces a −0.9…−1.5 mag/decade SQM-vs-exposure slope and kills short exposures outright ("background unresolved").

The `BlackLevelTracker` already measures the in-session intercept but was suppressed whenever a wizard calibration existed — backwards, since a stored bias is one moment's measurement of a moving level. This flips the priority in both SQM paths: a confidently fitted tracked intercept wins for bias; the wizard's dark-current rate stays authoritative on top. Replayed against the imx296 sweeps the tracker conditions on all four (stderr 0.26–0.32), fits 55.9–56.3, and the exposure slopes collapse to ≈0.

## 3. Wizard capture fixes (`fix(sqm): wizard captures gate on actual exposure...`)

The wizard's own calibration reports show why it measured 58 on a 56 sensor:
- the first ~3 frames of every stage report `actual_exposure_us` at the **previous** exposure (`req=1µs, actual=999999µs`) yet pass the frame gate, which checks the `exposure_time` field (echoes the committed setting, not what the sensor delivered);
- after the change lands, the OB clamp re-settles over ~3 more frames (imx296: 56→59 ADU) — and the settled clamp is genuinely exposure-dependent (~56 @ 1s vs ~59 @ 29µs min), so a min-exposure "bias" is only valid for its own regime (which is why fix 2 takes priority in operation).

Fix: gate on `actual_exposure_us` with a re-capture on stale frames, and discard clamp-settle frames after every exposure change in the bias and dark stages (~30–40s added to a wizard run).

## Cross-device validation

Full pipeline replayed offline over every referenced sweep (cloud-affected nights excluded):

| sensor | device(s) | sweeps | median err vs reference | spread |
|---|---|---|---|---|
| imx462 | mrosseel mr2 | 12 | −0.01 | ±0.15 |
| imx462 | rich | 3 | −0.21 | ±0.13 |
| imx296 | rich | 4 | −0.43* | ±0.10 |
| hq | mrosseel mr | 4 | +0.17* | ±0.20 |

\* constant, absorbed by a one-time `sqm_band_offset` refit for the new pipeline (imx296 −0.22→+0.21, hq 0.60→0.43) — deliberately **not** in this PR so the code change and the recalibration can be reviewed separately. Exposure-slope |mean| dropped from 1.20 to ~0.3 mag/decade on the imx296 and usable frames per sweep roughly doubled.

Validated: 141 tests across the SQM suites (18 new/updated); ruff + mypy clean on touched files.

https://claude.ai/code/session_01PXDxJkU5CzAegv3FcMEwEq

### DIFF
--- a/docs/ax/camera/CONTEXT.md
+++ b/docs/ax/camera/CONTEXT.md
@@ -58,6 +58,36 @@ The number of consecutive zero-match solve attempts required before recovery act
 Zero-match recovery was briefly a plugin point with four selectable strategies (Sweep, Exponential, Reset, Histogram) behind the Experimental "AE Algo" menu. [ADR 0010](../../adr/0010-zero-match-recovery-single-ladder.md) kept the Sweep ladder as the only behavior and removed the rest, the plugin seam, the menu, and the `auto_exposure_zero_star_handler` config key. Recovery is now the single concrete `ZeroMatchRecovery` class.
 _Avoid_: AE algo, zero-star handler, handler, plugin.
 
+### Frame extents
+
+Two regions of the sensor are in play at once, and confusing them is the
+standing hazard in this area. Always say which one you mean.
+
+**Full-sensor frame**:
+Every pixel the sensor delivers, uncropped. Written by `capture_raw_file()` —
+always, with no option to write anything narrower — and archived by the
+exposure sweep capture as `*_rawfull_*.tiff`. Nothing measures it: it exists so
+the margins are on disk for later analysis, because a margin not captured is
+gone for good.
+_Avoid_: raw frame (unqualified — the crop is equally raw).
+
+**Crop**:
+The centred square region of the sensor. This is what `cam_raw()` holds, what
+SQM photometry measures, and what each sweep frame's `raw_stats` covers. The
+crop is a plain slice of the full-sensor frame, so
+`CameraProfile.ensure_cropped()` recovers it exactly from an archived
+full-sensor frame — that equivalence is what lets full-sensor sweeps reproduce
+every number the cropped-era archive produced.
+
+The asymmetry inside a single sweep frame is deliberate and worth stating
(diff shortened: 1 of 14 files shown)


brickbots:
Thanks for continuing to work to make the SQM system even better @mrosseel I'm going to merge this to start the release process for 2.6.1

PR #548, opened 2026-07-25
Survive logind RemoveIPC: recover when the cedar shmem segment vanishes

brickbots:
## The bug

systemd-logind's default `RemoveIPC=yes` deletes every POSIX shared-memory segment a user owns the moment that user's **last login session ends**. A plain SSH logout is enough — the PiFinder services run as `pifinder` but hold no login session of their own. When that happens, `/dev/shm/cedar_detect_image` vanishes under the running solver:

- the cedar-detect server can't open the segment by name → gRPC `INTERNAL`
- `PFCedarDetectClient.extract_centroids`'s recovery path calls the upstream `_del_shmem()`, whose `unlink()` raises `FileNotFoundError` **before** `_use_shmem = False` is reached
- the designed fallback (image inlined in the gRPC request) never engages, and **every subsequent solve repeats the crash until the app restarts**

Any field user who SSHes into their unit and logs out silently loses plate solving.

## How it was found

The battery-runtime bench campaign: all four discharge runs across both campaigns "lost camera solving at the cable pull" — which turned out to be the operator's SSH logout seconds beside the pull. Device logs show ~28,000 identical solver exceptions per night (`FileNotFoundError: '//cedar_detect_image'` + server `errno 2`), first occurrence 3 s apart on two independent units. This also retro-explains campaign 1's degraded runs, which had been attributed to IMU pseudo-motion blanking.

## The fix (three layers)

- **`PFCedarDetectClient._del_shmem` override**: a segment already gone from `/dev/shm` is the method's goal state — treat it as released. The same `extract_centroids` call then retries with the image inlined, so even the frame that hits the error still gets centroids.
- **`pifinder_setup.sh`** (fresh installs): drop `RemoveIPC=no` into `/etc/systemd/logind.conf.d/` so provisioned units keep the fast shared-memory handoff instead of falling back after the first logout.
- **Migration `v2.6.1`** (existing units): delivers the same logind drop-in on the next software update via the standard one-shot mechanism in `pifinder_post_update.sh`, with a `try-restart` of logind so it applies without waiting for a reboot (logind restart does not end sessions, and RemoveIPC only acts at session end). *Naming note: the migration marker is just a label — if the next release ships under a different number, feel free to rename `v2.6.1` before merge.*

## Tests

`python/tests/test_solver_cedar_client.py`:
- vanished-segment release doesn't raise; no-segment release is a no-op
- end-to-end fallback: real shm segment created, unlinked out from under the client (exactly what logind does), faked gRPC stub returns `INTERNAL` then succeeds — asserts recovery in the same call, `_use_shmem` disabled, pixels inlined in the retry

Verified the test guards a real regression: with `_del_shmem` rebound to the upstream implementation the same scenario raises `FileNotFoundError` and never falls back.

Full suite: 991 unit+smoke passed; ruff + mypy clean; both shell scripts pass `bash -n`.

### DIFF
--- a/migration_source/v2.6.1.sh
+++ b/migration_source/v2.6.1.sh
@@ -1,0 +1,14 @@
+# Keep POSIX shared memory alive across SSH logouts: logind's default
+# RemoveIPC=yes deletes all IPC owned by the pifinder user (including the
+# solver's cedar-detect /dev/shm segment) the moment that user's last login
+# session ends — the PiFinder services run as pifinder but hold no login
+# session of their own, so a plain SSH logout used to degrade solving until
+# restart. The solver also recovers in software (PFCedarDetectClient), but
+# this keeps the fast shared-memory handoff in place. Mirrors the same
+# drop-in written by pifinder_setup.sh for fresh installs.
+sudo mkdir -p /etc/systemd/logind.conf.d
+printf '[Login]\nRemoveIPC=no\n' | sudo tee /etc/systemd/logind.conf.d/pifinder-removeipc.conf
+# Apply without waiting for a reboot; harmless if logind isn't running.
+# (Restarting logind does not end sessions, and RemoveIPC only acts at
+# session end, so nothing is reaped by the restart itself.)
+sudo systemctl try-restart systemd-logind.service || true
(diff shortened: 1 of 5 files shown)


mrosseel:
can you asks your agent if this is also an issue in nixos? Some cedar protection has happened there but not sure if it covers this

brickbots:
@mrosseel I believe that deleting the shared memory segments on logout is a standard logind behavior, so will probably also happen on NixOS, but the same fix should apply for NixOS if it uses logind/systemd :+1:

PR #553, opened 2026-07-28
Exposure sweeps archive the full sensor, not the square crop

mrosseel:
Sweeps saved the same square crop that photometry measures, so the sensor margins were discarded at capture. Every archived sweep is cropped, which makes questions about the margins — how far illumination falls off out there, whether any usable optical black exists — unanswerable from the frames we already have.

The crop is a plain slice, so the full frame is a **superset**: cropping later is exact and free, while margins not captured now are gone for good.

## What changes

| | Before | After |
|---|---|---|
| Sweep TIFF on disk | square crop | **full sensor**, named `*_rawfull_*.tiff` |
| Live SQM photometry (`cam_raw()`) | crop | **crop — unchanged** |
| Per-frame `raw_stats` | crop | **crop — unchanged** |
| Existing calibration constants | | **unchanged** |

`capture_raw_file()` always writes the whole sensor, with no option to write the crop instead. Readers reduce it with `CameraProfile.ensure_cropped()`, which routes through the ordinary `crop_and_rotate()`, so a replayed full-sensor sweep reproduces the cropped era's numbers *by construction* rather than by a parallel reimplementation. Tests pin that equivalence element for element.

## Two things deliberately stay on the crop

- **Live photometry** reads `cam_raw()`, untouched, so existing calibration stays valid.
- **`raw_stats`** stay crop-only. They are the black-level-versus-temperature series, and the vignetted margins would shift every mean and percentile, silently ending comparability with the existing archive. The TIFF holds everything, so full-sensor statistics remain computable later.

That asymmetry — full-sensor TIFF, crop-only statistics beside it — is easy to forget and expensive to get wrong, so it is named rather than implied:

- `CameraPI.last_cropped_frame`, `sweep_frame_record(cropped_frame=...)`
- `raw_stats.extent: "crop"` in each frame record
- `camera.raw_frame_extent: "full_sensor"` in `sweep_metadata.json`
- a **Frame extents** section in `docs/ax/camera/CONTEXT.md` defining *full-sensor frame* vs *crop*, cross-referenced from SQM's *raw photometry image*

`raw_stats` keeps its key name: existing archives use it, and forking the schema is the thing this change exists to avoid. Records from before this change simply lack `extent` — correctly, since back then the TIFF was the crop too.

## Compatibility

Full-sensor frames are named `rawfull` so they fall **outside** the glob readers use for the cropped era's `*_raw_RGGB.tiff`. An un-updated reader finds nothing rather than silently scaling centroids against the wrong frame size — a loud failure instead of a quiet wrong answer.

## Testing

`ruff` clean, 1002 unit+smoke tests pass. New coverage: full-sensor/crop era detection, exact reduction to the live crop across all four sensor profiles, cropped-era passthrough, and that `raw_stats` cover the crop when the sibling TIFF is full-sensor.

### DIFF
--- a/docs/ax/camera/CONTEXT.md
+++ b/docs/ax/camera/CONTEXT.md
@@ -58,6 +58,36 @@ The number of consecutive zero-match solve attempts required before recovery act
 Zero-match recovery was briefly a plugin point with four selectable strategies (Sweep, Exponential, Reset, Histogram) behind the Experimental "AE Algo" menu. [ADR 0010](../../adr/0010-zero-match-recovery-single-ladder.md) kept the Sweep ladder as the only behavior and removed the rest, the plugin seam, the menu, and the `auto_exposure_zero_star_handler` config key. Recovery is now the single concrete `ZeroMatchRecovery` class.
 _Avoid_: AE algo, zero-star handler, handler, plugin.
 
+### Frame extents
+
+Two regions of the sensor are in play at once, and confusing them is the
+standing hazard in this area. Always say which one you mean.
+
+**Full-sensor frame**:
+Every pixel the sensor delivers, uncropped. Written by `capture_raw_file()` —
+always, with no option to write anything narrower — and archived by the
+exposure sweep capture as `*_rawfull_*.tiff`. Nothing measures it: it exists so
+the margins are on disk for later analysis, because a margin not captured is
+gone for good.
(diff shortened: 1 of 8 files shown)


mrosseel:
Wrong home for this — it belongs in #544, which is the SQM sweep/photometry work. Folding it in there.

PR #560, opened 2026-07-31
SQM: sky-colour-dependent radiometric zero point + band-offset refit (follow-up to #544)

mrosseel:
Follow-up to #544, which merged the pipeline change but deferred the recalibration it needs — spotted while prepping the 2.6.1 notes. Re-deriving from the sweep archive found the deferred refit was the *smaller* half of the problem.

## The main finding: one constant can't fit both sky regimes

Over **23 referenced imx462 sweeps spanning 17.5–20.9 mag**, the shipped `radiometric_zero_point=15.25` reads **~0.10 mag dark** at the LP site and **~0.85 mag bright** at a dark one. Every attempt to fit a single constant kept disagreeing with itself, because it was averaging two regimes.

The cause is physical. The radiometer measures sky in the **sensor's** passband; the reference meter measures **V**. Converting between them depends on the sky's spectrum — light pollution is sodium/LED and green-weighted, airglow is grey and NIR-rich, which a bare Bayer sensor sees and a V-band meter does not. Sky brightness is only a proxy for that. **Sky colour measures it directly**, and it's already in the frame: measured R/G runs **0.83–0.89** at the LP site and **1.00–1.04** at the dark site.

| model | residual sd |
|---|---|
| constant (shipped) | 0.337 |
| vs sky brightness | 0.185 |
| **vs measured sky colour (R/G)** | **0.079** |

## Why I believe it isn't just a spare parameter

**Leave-one-night-out CV: MAE 0.247 → 0.108.** The load-bearing row is holding out `20260720`, the only dark night — trained on nothing but LP data, the model predicts an unseen regime at **0.312** against the constant's **0.944**. It extrapolates, rather than interpolating between fitted points.

**The same fit on the HQ is rejected by CV** (0.234 with colour vs 0.182 constant), and its colour slope measures **20× smaller** (+0.272 vs +5.544). That is exactly what a NIR-leak term must do on a sensor with a factory IR-cut filter. A model that wins everywhere would be suspicious; one that wins only where the physics says it should is evidence.

## Result against the archive

| | before | after |
|---|---|---|
| imx462 median residual | −0.045 | **+0.005** |
| imx462 spread | −0.23 … **+1.08** | −0.18 … **+0.16** |
| the three dark-site sweeps | +0.76, +0.77, +1.08 | +0.16, −0.02, +0.005 |
| hq median residual | +0.181 | **+0.000** |

The ~1 mag dark-site error is gone and the LP regime got *tighter*, not traded away.

## Shape of the change

- `collect_radiometer_sample` now records `background_red` / `background_green` off the mosaic (colour sensors only).
- `CameraProfile` gains `radiometric_colour_slope` / `_pivot` / `_range`. **Slope 0 is a plain constant**, so mono and IR-cut sensors are unchanged in behaviour.
- imx462/imx290: `15.25` → `15.159 + 5.544·(R/G − 0.85)`, R/G clamped to the calibrated 0.83–1.04 rather than extrapolated.
- hq: constant, `14.79` → `14.971`.
- A frame with no colour falls back to the constant, pinned at the LP-regime value.

## Also: the band offsets #544 promised (diagnostic only)

`sqm_band_offset` is read **only** by `sqm.py`, the stellar path — the published SQM comes from the radiometer, so these affect a diagnostic, not what users see.

#544's text quoted imx296 −0.22 → **+0.21** and hq 0.60 → **0.43**. Neither reproduces:

| profile | shipped | #544 text | re-derived | evidence |
|---|---|---|---|---|
| imx296 | −0.22 | +0.21 | **−0.02** | 4 sweeps, residuals +0.13…+0.22 — tight |
| hq | 0.60 | 0.43 | **0.99** | 9 sweeps, 0.67 mag scatter → 0.99 ± 0.2 |
| imx462 | 0.53 | — | 0.514 | **control** — validates the method |

The HQ band offset argues with the physics: it's nominally a passband term and the IR-cut implies ≈0, yet 0 puts the stellar SQM a full magnitude bright. So roughly a magnitude of the HQ *stellar* chain is unaccounted for and this constant is absorbing it. The comment says so, so whoever finds the real error refits rather than assuming the number transfers.

## Caveats, kept visible

- **imx296** rests on 4 sweeps from one night and one observer, and being **mono it cannot use the colour model at all** — its radiometric zero point is deliberately left untouched. A dark-site sweep from that camera would be the single most useful thing to add.
- The imx462 dark-site anchor is **one night** (3 sweeps). The CV result is what makes it credible, not the sample size.
- Sky colour is a proxy for spectrum, not a measurement of it; two different spectra with the same R/G would be conflated.

### DIFF
--- a/docs/adr/0026-radiometric-zero-point-keyed-to-sky-colour.md
+++ b/docs/adr/0026-radiometric-zero-point-keyed-to-sky-colour.md
@@ -1,0 +1,179 @@
+# SQM: radiometric zero point keyed to measured sky colour
+
+ADR 0022 made the radiometer the published SQM source, calibrated by one
+constant per sensor: `radiometric_zero_point`. Re-deriving that constant over
+23 referenced imx462 sweeps spanning 17.5–20.9 mag skies, every attempt to fit
+a single value disagreed with itself. The shipped `15.25` read about **0.10 mag
+dark** at the light-polluted reference site and about **0.85 mag bright** at a
+dark one. Averaging the two regimes produced a constant that was wrong at both
+ends, and no amount of refitting fixed that, because the thing being fitted was
+not a constant.
+
+The cause is physical rather than statistical. The radiometer measures sky in
+the **sensor's** passband; the reference meter measures **V**. The conversion
+between them depends on the sky's *spectrum*: light pollution is sodium/LED and
+green-weighted, airglow is grey and NIR-rich, and a bare Bayer sensor sees that
+NIR while a V-band meter does not. Sky brightness is only a proxy for spectrum —
+a bright airglow sky and a dim urban sky are not the same colour. **Sky colour
+measures it directly**, and it is already in the frame: measured R/G runs
+0.83–0.89 at the light-polluted site and 1.00–1.04 at the dark one.
+
+| model for the zero point | residual sd |
+|---|---|
+| constant (previously shipped) | 0.337 |
+| linear in sky brightness | 0.185 |
+| **linear in measured sky colour (R/G)** | **0.079** |
+
+## The decision
+
+**The radiometric zero point is a function of measured sky colour, not a
+constant.**
+
+```text
+effective_zero_point = radiometric_zero_point
+                       + radiometric_colour_slope × (clamp(R/G) − radiometric_colour_pivot)
+```
+
+1. **`radiometric_colour_slope = 0` is a plain constant**, and is the default.
+   Mono sensors have no colour to measure; a factory IR-cut sensor has
+   essentially no NIR leak to correct. Both keep the previous behaviour exactly,
+   so this change is scoped to the sensors whose physics calls for it — imx462
+   and imx290 (`5.544` mag per unit R/G, pivot `0.85`).
+
+2. **R/G is clamped to the calibrated range, never extrapolated.** The fit is
+   evidence about the colours it saw (`0.83–1.04`) and nothing else. Both the
+   raw and clamped ratios are recorded so an out-of-range site is visible in the
+   archive rather than silently absorbed.
+
+3. **The pivot is where the correction is zero**, chosen so a frame carrying no
+   colour falls back to a sensible light-pollution-regime constant rather than
+   to the fit's intercept. Mono sensors, and any frame whose mosaic phase cannot
+   be trusted, take that path.
+
+4. **`radiometric_zero_point` keeps meaning the profile constant.** The value
+   actually applied is reported separately as
+   `radiometric_zero_point_effective`, always present whether or not a
+   correction was made, so archives stay comparable across this change.
+
+5. **Mosaic phase is checked, not assumed.** Reading the red plane is more
+   fragile than reading green, and the difference is not obvious: a 180°
+   rotation maps the block `R G / G B` to `B G / G R`, so the two green sites
+   are invariant while red and blue swap. An odd crop origin does the same.
+   `crop_and_rotate` runs before the radiometer, so both are reachable from
+   profile configuration alone. `_mosaic_phase_is_rggb` requires RGGB order (not
+   merely "some Bayer format"), zero rotation, and even crop origins, and
+   reports *no colour* rather than a wrong colour. A wrong R/G does not fail
+   loudly — it returns a plausible number that is up to the clamp width wrong.
+
+## Why this is not just a spare parameter
+
+A free parameter always reduces in-sample scatter, so in-sample residuals cannot
+justify one. Two things do.
+
+**Leave-one-night-out cross-validation: MAE 0.247 → 0.108.** The load-bearing
+row is holding out `20260720`, the only dark night. Trained on light-polluted
+data alone, the colour model predicts an unseen regime at **0.312** against the
+constant's **0.944**. It extrapolates rather than interpolating between fitted
+points.
+
+**The same fit on the HQ is rejected by cross-validation** (0.234 with colour vs
+0.182 constant), and its colour slope measures **20× smaller** (+0.272 vs
++5.544). That is precisely what a NIR-leak term must do on a sensor with a
+factory IR-cut filter. A model that won everywhere would be suspicious; one that
+wins only where the physics predicts it should is a negative control, and it is
+the strongest evidence here.
+
+Against the archive, after the change:
+
+| | before | after |
+|---|---|---|
+| imx462 median residual | −0.045 | **+0.005** |
+| imx462 spread | −0.23 … **+1.08** | −0.18 … **+0.16** |
+| the three dark-site sweeps | +0.76, +0.77, +1.08 | +0.16, −0.02, +0.005 |
+| hq median residual | +0.181 | **+0.000** |
+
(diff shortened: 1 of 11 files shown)


brickbots:
## Review + follow-up commit

Reviewed this and pushed one commit to the branch (`212e5f24`, +1008/−27 over 10 files) with the review follow-ups applied. **No changes to the shipped constants or to the model itself** — the calibration work stands exactly as you derived it. Everything below is guardrails, coverage, tooling and docs.

Before the findings: this is unusually careful calibration work. The thing that convinced me is the **negative control** — a model that won everywhere would be a spare parameter, and one that CV *rejects* on the IR-cut HQ with a 20× smaller slope is behaving the way a NIR-leak term must. Leading with that is right. The archive discipline is also easy to get wrong and you didn't: keeping `radiometric_zero_point` meaning the profile constant and reporting `_effective` alongside it is what keeps pre- and post-change archives comparable, and there's a test pinning it.

Verification on the branch as it now stands: **1105 unit+smoke tests pass** (17 new), `ruff check` and `ruff format --check` clean, `mypy` clean on the changed SQM modules.

### 1. Guarded the Bayer-phase invariants

`_sky_red_green` gated on `format.startswith("S")` and then hardcoded RGGB positions, while `extract_photometry_image` twelve lines above required `startswith("SRGGB")`.

This matters more than the inconsistency suggests, because **red is far more phase-fragile than green**. A 180° rotation maps `R G / G B` to `B G / G R` — the green sites are invariant, red and blue swap. An odd crop origin does the same. `crop_and_rotate` runs before the radiometer, so both are reachable from profile config alone. Verified on a synthetic frame (true R/G 0.850, blue/G 0.500):

| input | R/G returned | green plane |
|---|---|---|
| as-is | **0.850** ✓ | 1000 |
| rotated 180° | **0.500** ✗ (blue read as red) | 1000 |
| odd crop origin | **0.500** ✗ | 1000 |

**This is correct as shipped** — all three colour profiles are `rotation_90=0` with even crops (462/290 `(50,50)`/`(470,470)`, hq `(0,0)`/`(256,256)`), so it was latent, not live. But it fails *silently* into a plausible number. New `_mosaic_phase_is_rggb` requires RGGB order, zero rotation and even crop origins, and reports no colour rather than wrong colour. Behaviour today is unchanged; a future orientation change now fails a test instead of quietly biasing everyone's SQM.

Credit where due: the clamp bounds the blast radius of any such error to roughly ±0.5 mag rather than unbounded, which is exactly what defensive clamping is for.

### 2. Covered the mosaic extraction path

The four colour tests all hand-build `background_red`/`background_green` and start at `radiometric_sqm`, so `_sky_red_green` — the fragile half — was never executed by the suite. Added: phase correctness at three frame sizes (including the odd-border path, which your `by += by % 2` handles correctly — I checked 215×215 → border 21), guard rejection across five broken profile variants, a profile-level assertion that the shipped three hold the invariants, mono returning no colour, an end-to-end mosaic→corrected-value round trip, and one test that pins the blue-as-red failure itself so the guard's rationale stays checkable rather than living only in a comment.

### 3. Sweep metadata records the whole model

That block's comment says it exists so a sweep is "self-describing", and it recorded `bias_offset` / `sqm_band_offset` / `color_coefficient` but not the radiometric zero point. This PR is what makes the gap bite — the zero point is no longer a constant you can look up by camera type — so sweeps archived from here on now record slope, pivot, range and field width too. That's precisely what the *next* refit needs.

### 4. Made the derivation reproducible in-repo

`evaluate_radiometer_archive.py` scored sweeps with the colour term applied but still fitted only a single constant, so rerunning it produced a `fitted_radiometric_zero_points` value contradicting the shipped model with nothing explaining why. None of the PR's evidence — the colour regression, the LONO CV, the HQ rejection — was reproducible from the repo.

It now emits a `radiometric_models` block that re-derives both models, cross-validates them, and reports a verdict plus an `agrees_with_shipped` flag. New `PiFinder/sqm/radiometric_fit.py` holds the maths, unit tested on synthetic sweeps with known answers (including the negative control: colour must be rejected when the sensor doesn't care, *even though it wins in-sample*). Two deliberate choices, both documented in the module:

- **Per sweep, not per frame** — frames in a sweep share a sky, observer and reference reading; pooling lets a long sweep outvote a whole night.
- **Held-out error, not in-sample scatter** — least squares can't do worse in-sample than the no-slope fit, so that comparison can never reject a colour term however useless it is.

### 5. Docs + ADR 0026

`docs/ax/sqm.md` still carried the constant-only formula and named all three defaults at their pre-PR values. `CONTEXT.md` gains **sky colour (R/G)**, **radiometric colour term** and **mosaic phase** as glossary terms, and the *Radiometric SQM* entry no longer calls it the "fixed-calibration" value.

**ADR 0026** carries the decision, the evidence, the rejected alternatives (constant refit, keying to sky brightness, colour term on every sensor, extrapolating past the range, measuring spectrum properly), and the caveats. I moved the **one-dark-night** limitation into the ADR and pointed the profile comment at it — your PR text is honest about it, but the code comment leads with the CV numbers alone, and the code is what the next maintainer reads. The slope is set by the lever arm between two colour clusters with essentially nothing between them, so a systematic peculiar to that night maps straight onto it.

---

## One question back: which CV number is `0.312`?

Writing the cross-validation surfaced something worth pinning down. When LONO holds out the only night of a regime, clamping to the **training** colour range pins the prediction at the edge of the fitted span — so the model **cannot extrapolate at all**, and most of the theoretical gain is unavailable.

That means "would shipping this model have helped?" and "does the physical relation extrapolate?" are different questions with different answers. The tool now reports both (`colour_mae` clamped, `colour_mae_unclamped`) and flags nights that fell outside the training range; the verdict uses the clamped figure as the conservative one.

Do you recall whether the `0.944 → 0.312` figure was clamped or unclamped? It doesn't change the decision either way — but the practical consequence is real: at a site outside `0.83–1.04`, the clamp deliberately under-corrects rather than trusting the fit off its own end, and that's worth being explicit about.

## Left alone

- **`release-2.6.1-test-plan.md` line 225** (open decision P1.3, quoting "imx296 ~0.4 low, HQ ~0.2 high") — that file was added to `main` after this branch forked, so it isn't reachable from here. Needs a one-line update at merge time; I didn't want to rebase or merge `main` into your branch just for that.
- **HQ `sqm_band_offset` 0.60 → 0.99.** Not an objection — the comment is refreshingly candid that it fights the physics, and blast radius is genuinely low since it's read only by the diagnostic stellar path. Flagging only that zeroing the median of a chain you believe is broken also removes the signal that something *is* broken. Your call; the comment discharges the duty to warn.
- Minor nits from the review left as-is: stale imx296 provenance comment above `sqm_band_offset=-0.02` (the "single moonlit 2025-10-31 sweep" lines describe the old −0.22), and `test_sqm.py:1832` commenting `0.43` where line 1846 asserts `0.53` (pre-existing).

Happy to drop the follow-up commit if you'd rather apply any of it differently — it's a clean single commit on top of yours.

<sub>Review and follow-up commit produced with [Claude Code](https://claude.com/claude-code).</sub>

PR #609, opened 2026-08-14
Derive field of view from the optical train (sensor × lens)

brickbots:
Implements [ADR 0027](../blob/worktree-lens-config-impl/docs/adr/0027-fov-gate-derived-from-optical-train.md). The first commit is the docs/vocabulary work; the second is the implementation.

## Why

Field of view was three hard-coded constants that disagreed with each other and with the hardware:

| consumer | constant |
|---|---|
| solver | `fov_estimate=12.0`, `fov_max_error=4.0` |
| SQM | `radiometric_fov_degrees` per sensor (13.71 / 10.38 / 10.34) |
| chart frustum | `9.5` |

The solver pair covered three of the four sensor × lens combinations we can ship, and nothing said which one it was for. A 12 mm lens on an imx296 images 17.8°, outside the `[8.0, 16.0]` window — tetra3 enforces that window twice, so such a frame does not solve **at all**, however many good centroids it has.

## What changed

- **`PiFinder.optics`** (new) — `Lens` (nominal focal length for the menu label, **effective** for all arithmetic, plus f-number), a three-entry registry, and `OpticalTrain` exposing `fov_degrees`, plate scale against a *named* pixel grid, and `solver_fov_params()`.
- **`CameraProfile` relocated** out of `sqm/` to `PiFinder.camera_profiles` — it describes the sensor, and SQM is only one of several consumers. Gains `pixel_pitch_um` and `default_lens_key`; `radiometric_fov_degrees` is removed and derived instead. `sqm/camera_profiles.py` re-exports for existing callers.
- **FOV gate = ±15%** of the derived field of view, resolved per frame and memoized on `(camera_type, camera_lens)`, so a lens change lands on the next frame rather than the next boot. Resolved lazily rather than at `solver()` init — `camera_type` still holds the pre-camera default at startup.
- **Settings → Advanced → Lens**, beside Camera Type. `UITextMenu` now lets a `value_callback` supply the default when a config option is unset, so the menu shows the lens actually in force rather than its first entry.
- **Frustum shading** on the align screen and the web chart API follows the same derived value.

## Zero migration

An install with no `camera_lens` key resolves to the sensor's shipped lens (hq → 25 mm, imx296/imx462 → 16 mm). That reproduces the previously stored field widths to within 0.03° — under 0.01 mag of SQM, so **no existing user's reading moves**.

Deriving from the nominal 16.0 rather than the measured 15.61 would have shifted every user's SQM by ~0.05 mag. `tests/test_optics.py` pins both the reproduction and that margin; 15.61 reproducing *two* independently calibrated sensors is what makes it a measurement of the lens rather than a fudge factor.

## Debug camera relabel (load-bearing)

`camera_debug.py` now declares `hq` rather than `imx296`. The frames in `test_images/` are ~10.2°, which only the hq train's gate covers — without this, `--camera debug` stops solving entirely. `tests/test_optics_solving.py` runs the real frames through tetra3 and guards both directions.

## Verification

- Full suite: **1478 passed**, 8 failures all pre-existing (confirmed identical on a clean `origin/main` worktree — `test_multiproclogging`, `test_radec_entry`, `test_ui_modules`). `ruff check` and `mypy` clean.
- 38 new tests across derivation, gate, plate scale, zero-migration, live re-read, and real-frame solving.
- **On the running app** (`--camera debug`, headless): solves land (21 matches, fitted FOV 10.20° vs derived 10.33°); the Lens menu shows **25mm** checked with no config entry; selecting 16mm stops solving immediately with no restart; selecting 25mm resumes it.

## Deliberately not done

**No recovery from a mis-stated lens** — the gate is centred on what the user said is fitted, so a wrong statement puts every frame outside it. Self-healing would add a recovery path through the solver's hot loop for a once-per-device setting; the lens is marked on the barrel, so this documents cleanly. The sharp edge is that it *presents as an exposure problem*, because auto-exposure is driven by `Matches` rather than centroid count. Mitigated here by logging the resolved train on change and erroring when the derived gate falls outside the pattern database's range.

## Follow-ups (filed, not in this PR)

- **#610** — Auto-exposure is driven by `Matches`, so a rejected frame looks underexposed. `SolveDiagnostics` has no centroid count, so a solver-side rejection is indistinguishable from a dark frame.
- **#611** — Gate zero-match recovery on centroid count. Shares #610's prerequisite (`Centroids` on `SolveDiagnostics`) and fixes the pre-existing defocus thrash too.
- **#612** — Calibrate the 12 mm lens on-sky. Its effective focal length is the nominal 12.0 standing in for a measurement (flagged in the data as `effective_focal_length_measured=False`), its SQM zero-point offset is unmeasured, and its f-number is the conservative f/2.0. One sweep closes all three.
- **#613** — Docs: "changed lens → no solves" troubleshooting path, plus `docs/source/BOM.rst:96` still advises "something fast with a 10deg FOV", which is imx462-specific and contradicts the parts list two lines above it.


brickbots:
Review follow-ups folded in as three commits on this branch. Each is one of the items picked out of the review; nothing else from it is included.

## 1. `save_sweep_metadata(lens_key=...)` was never passed

`camera_interface.py`'s sweep-completion block passed `camera_type` but not `lens_key`, so a sweep captured on a device with a non-default lens recorded the *shipped* lens's field width and a wrong `"lens"` string. Those archives are what `evaluate_radiometer_archive.py` re-derives the SQM constants from, so the mislabelling is self-reinforcing. Fixed, plus the now-false docstring ("omitted means the sensor's shipped lens, which is what the device itself assumes" — the device now assumes the *configured* lens), plus regression tests that a stated lens moves `radiometric_fov_degrees` off the shipped value.

## 2. Unknown `camera_type` no longer raises

`get_camera_profile()` raises for anything outside `CAMERA_PROFILES`, and this PR put that lookup on three paths that were previously insensitive to `camera_type` — the solver's per-frame resolve, the align screen, and the Lens menu's value callback. In the solver a `ValueError` escapes the local handler and reaches the outer one that restarts the entire solver loop, tetra3 database load included: a crash-reload loop, once per frame.

`optics.resolve_camera_profile()` is now the single fallback policy for the sensor half — the counterpart of `resolve_lens` for the lens half. It logs and falls back to `imx296`, which is `SharedStateObj`'s pre-camera default, so an unknown sensor behaves like the window every boot already passes through. `get_camera_profile` stays strict so scripts and tests still hear about a typo. `ui/callbacks.get_camera_lens`, which bypasses the resolver, goes through the same helper.

**Hardening, not a fix for an observed crash.** `camera_none` does publish `camera_type="none"`, but its `get_images` passes a `bias_image` argument `CameraInterface.get_image_loop` does not accept, so the camera process dies before `"none"` ever reaches shared state. That `TypeError` predates this PR and is unrelated — worth its own issue rather than folding in here.

## 3. `plot.py` frustum / `camera_fov` default — Option A + C

- `DEFAULT_CAMERA_FOV_DEGREES` is gone. It was a hard-coded constant wearing a derived costume: it did not become correct, it moved from 9.5° to 13.71°, and spelling it as a derivation made it *harder* to notice it was a constant.
- `camera_fov=None` now means no frustum — no shading **and** no `visible_stars` filter. The filter used to run ungated on `shade_frustrum`, so a caller could have its stars silently restricted by a value it never set, with no way to ask for "chart, no frustum".
- `frustum_box()` holds the geometry, so shading and filter read the same box and "no frustum" is one testable state instead of two inequalities.
- Passing it per render also takes it off the shared object: `api_extensions` caches one `Starfield` across requests and waitress is multi-threaded, so `set_camera_fov()` between one request's resolve and its render belonged to whoever got there last.

`align.py` and the chart API set it as before → behaviour unchanged. `ui/chart.py`, which never set it, now gets unfiltered `visible_stars` that it already discards. `docs/ax/camera/CONTEXT.md` gains a **Frustum** entry, since "no frustum" is now a real state and `visible_stars` means *stars inside the frustum*, not *stars on the chart*.

## Verification

- `pytest -m "unit or smoke"` — 1209 passed, including optics + optics_solving (44), sqm/radiometer/solver_sqm/sweep_frame_record (165), menu_struct/chart/align/api/camera_interface (51).
- `ruff check`, `ruff format --check`, `mypy` on the touched modules — clean.
- Driven on the running app (debug camera → `hq`): the Lens menu opens with 25mm ticked; align draws **no** frustum at its default 10.2° zoom because an hq device images 10.33° — same as before — and zooming to 20° draws the box at the computed offsets (42.6–133.4 px on a 176 px display), surround dimmed to 32, interior 0.

## Deliberately not done

- The Lens menu offering combinations below the pattern database's `min_fov`; the existing `_warn_if_outside_solver_database` log stands.
- The `solver.py` double `camera_type()` read, the `.po`/offline-script nits, the hq non-square-crop rounding note.
- ADR 0027's consequences section could use one sentence noting `--camera debug` now reports `"hq"` to *every* `camera_type()` consumer (SQM profile, API status, sweep metadata), not just the FOV gate — all dev-only, no impact. Left out as outside the three items; happy to add it.
- Pre-existing and untouched: the cached API `Starfield` still carries per-render state (`projection`, `roll`) across concurrent requests. Only `camera_fov` moved off it.

PR #615, opened 2026-08-15
PR #609 follow-ups: sweep lens, unknown-sensor fallback, explicit frustum

brickbots:
Review follow-ups for #609, stacked on its branch (`worktree-lens-config-impl`) so #609 itself is untouched. Each commit is one of the three items Rich picked; nothing else from the review is in here.

If you would rather have these inside #609, this branch is a direct descendant of its tip, so `git push origin worktree-pr609-followups:worktree-lens-config-impl` is a clean fast-forward.

## 1. `save_sweep_metadata(lens_key=...)` was never passed

`camera_interface.py`'s sweep-completion block passed `camera_type` but not `lens_key`, so a sweep captured on a device with a non-default lens recorded the *shipped* lens's field width and a wrong `"lens"` string. Those archives are what `evaluate_radiometer_archive.py` re-derives the SQM constants from, so the mislabelling is self-reinforcing. Fixed, plus the now-false docstring ("omitted means the sensor's shipped lens, which is what the device itself assumes" — the device now assumes the *configured* lens) and a regression test that a stated lens moves `radiometric_fov_degrees` off the shipped value.

## 2. Unknown `camera_type` no longer raises

`get_camera_profile()` raises for anything outside `CAMERA_PROFILES`, and #609 put that lookup on three paths that were previously insensitive to `camera_type` — the solver's per-frame resolve, the align screen, and the Lens menu's value callback. In the solver a `ValueError` escapes the local handler and reaches the outer one that restarts the entire solver loop, tetra3 database load included: a crash-reload loop, once per frame.

`optics.resolve_camera_profile()` is now the single fallback policy for the sensor half — the counterpart of `resolve_lens` for the lens half. It logs and falls back to `imx296`, which is `SharedStateObj`'s pre-camera default, so an unknown sensor behaves like the window every boot already passes through. `get_camera_profile` stays strict so scripts and tests still hear about a typo. `ui/callbacks.get_camera_lens`, which bypasses the resolver, goes through the same helper.

**This is hardening, not a fix for an observed crash.** `camera_none` does publish `camera_type="none"`, but its `get_images` passes a `bias_image` argument `CameraInterface.get_image_loop` does not accept, so the camera process dies before `"none"` ever reaches shared state. That `TypeError` predates #609 and is unrelated — worth its own issue, not folded in here.

## 3. `plot.py` frustum / `camera_fov` default — Option A + C

Rich had not picked an approach; this implements the recommended **A** (delete the default, `None` means *no frustum*) together with **C** (pass it per render instead of holding it on the object), which the handoff called the honest end state.

- `DEFAULT_CAMERA_FOV_DEGREES` is gone. It was a hard-coded constant wearing a derived costume: it did not become correct, it moved from 9.5° to 13.71°, and spelling it as a derivation made it *harder* to notice it was a constant.
- `camera_fov=None` now means no frustum — no shading **and** no `visible_stars` filter. The filter used to run ungated on `shade_frustrum`, so a caller could have its stars silently restricted by a value it never set, with no way to ask for "chart, no frustum".
- `frustum_box()` holds the geometry, so the shading and the filter read the same box and "no frustum" is one testable state instead of two inequalities.
- Per-render also takes it off the shared object: `api_extensions` caches one `Starfield` across requests and waitress is multi-threaded, so `set_camera_fov()` between one request's resolve and its render belonged to whoever got there last.

`align.py` and the chart API set it as before → behaviour unchanged. `ui/chart.py`, which never set it, now gets unfiltered `visible_stars` that it already discards.

Vocabulary: `docs/ax/camera/CONTEXT.md` gains a **Frustum** entry, since "no frustum" is now a real state and `visible_stars` means *stars inside the frustum*, not *stars on the chart*.

## Verification

- `pytest -m "unit or smoke"` — 1209 passed (was 1205 + 4 new files' worth of cases), including the handoff's named suites: optics + optics_solving 44, sqm/radiometer/solver_sqm/sweep_frame_record 165, menu_struct/chart/align/api/camera_interface 51.
- `ruff check`, `ruff format --check`, `mypy` on the touched modules — clean.
- Driven on the running app (`pifinder-remote`, debug camera → `hq`): the Lens menu opens and shows 25mm ticked (item 2's callback path); the align screen draws **no** frustum at its default 10.2° zoom, because an hq device images 10.33° — same as before this change — and zooming to 20° draws the box at the computed offsets (42.6–133.4 px on a 176 px display) with the surround dimmed to 32 and the interior at 0.

## Not done (out of scope, per the handoff)

- The Lens menu offering combinations below the pattern database's `min_fov`; the existing `_warn_if_outside_solver_database` log stands.
- The `solver.py` double `camera_type()` read, the `.po`/offline-script nits, the hq non-square-crop rounding note.
- ADR 0027's consequences section could use one sentence noting that `--camera debug` now reports `"hq"` to *every* `camera_type()` consumer (SQM profile, API status, sweep metadata), not just the FOV gate — all dev-only, no impact. Left out as it is outside items 1–3; say the word and it is a one-line commit.
- Pre-existing and untouched: the cached API `Starfield` still carries per-render state (`projection`, `roll`) across concurrent requests. Only `camera_fov` moved off it.

### DIFF
--- a/docs/ax/camera/CONTEXT.md
+++ b/docs/ax/camera/CONTEXT.md
@@ -148,6 +148,20 @@ downscale factor and confusing them silently rescales any angle computed
 from it.
 _Avoid_: arcsec per pixel (unqualified — say which grid), resolution.
 
+**Frustum**:
+The box drawn on a chart marking the part of it the camera images — the
+**field of view** of the current optical train, inside the chart's own zoom
+level. It does two jobs at once: the chart outside it can be dimmed, and it
+selects the stars the chart reports back, which is why the align screen only
+offers alignment stars the camera can actually see. A chart is rendered with
+a frustum only when the caller states a camera field of view; **no frustum**
+is a real state (`plot.frustum_box` returns None), not a default box, because
+there is no stand-in for a number that is a property of the hardware fitted
+to that particular device.
+_Avoid_: camera box, FOV box, "the frustrum" in prose (the misspelling
+survives only in the `shade_frustrum` argument name); "visible stars" meaning
+*stars on the chart* — with a frustum they are the stars inside it.
+
 ### Cross-context terms
 
 - **`Matches`** — defined in [Positioning](../positioning/CONTEXT.md): count of stars tetra3 matched in the most recent solve attempt, published on every attempt (success or failure) because auto-exposure depends on it. The feedback signal for solver-driven auto-exposure.
--- a/python/PiFinder/api_extensions.py
+++ b/python/PiFinder/api_extensions.py
@@ -463,14 +463,14 @@ def register_api_routes(app, server_instance, require_auth=False):
             )
 
             # The frustum marks what the camera images, so it follows the
-            # fitted optical train. Set per request rather than baked into the
-            # cache key: the Starfield is expensive to build and is reused
-            # across requests, while the lens can change under it.
-            starfield.set_camera_fov(
-                _API_OPTICAL_TRAIN.resolve(
-                    ss.camera_type(), ss.camera_lens()
-                ).fov_degrees
-            )
+            # fitted optical train. Passed per render rather than held on the
+            # Starfield: that object is cached and shared across requests, and
+            # waitress serves them on several threads, so anything set on it
+            # between one request's resolve and its render belongs to whoever
+            # got there last.
+            camera_fov = _API_OPTICAL_TRAIN.resolve(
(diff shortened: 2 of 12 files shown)


brickbots:
Folded into #609 directly — `worktree-lens-config-impl` was fast-forwarded to these three commits, so GitHub marked this as merged. Write-up moved to a comment on #609.

Issue #613, opened 2026-08-15
Docs: "changed lens → no solves" troubleshooting, and BOM's FOV advice is sensor-specific


brickbots:
Two user-facing docs gaps left by #609 / ADR 0027.

## 1. Troubleshooting: "I changed the lens and now it never solves"

ADR 0027 makes the solver's FOV gate follow the lens the user says is fitted, and **deliberately does not recover from a wrong answer** — self-healing would add a recovery path through the solver's hot loop for a once-per-device setting. That decision is only defensible if the failure documents cleanly, which is currently doesn't exist.

What the page needs to say:

- The symptom is **no solves at all**, not degraded solving. A frame outside the gate is rejected however good it is.
- It **looks like an exposure problem** — the exposure ladder will be hunting — so the obvious diagnosis is the wrong one. (Tracked separately: #610, #611.)
- The fix is Settings → Advanced → **Lens**, and the lens is **marked on the barrel**, so the user can just read it.
- No restart needed: the solver re-reads the lens per frame and solving resumes on the next one.

Suggested home: the troubleshooting section of the user guide, cross-linked from wherever the Lens setting is documented. Worth a mention in the Lens menu documentation itself too, since that is where someone changing the setting will be looking.

The new menu also wants documenting in its own right — Settings → Advanced → Lens, three options (12mm / 16mm / 25mm), defaulting to whatever the device shipped with.

## 2. `docs/source/BOM.rst:96` gives imx462-specific advice

```rst
- Other lenses might work here, but something fast with a 10deg FOV is ideal
```

10° is right for a 16 mm lens on an **imx462** (10.40°). It is wrong for the other shipped trains:

| sensor | shipped lens | field of view |
|---|---|---|
| imx296 | 16 mm | 13.71° |
| imx462 / imx290 | 16 mm | 10.40° |
| hq | 25 mm | 10.33° |

Since the same BOM section says you can use either the imx296 or the imx462 module, the advice contradicts the parts list two lines above it. Better framing is probably the focal length (which is what the buyer actually selects) rather than a resulting FOV — the same 16 mm lens gives different fields on the two sensors.

Worth also noting the bundled tetra3 database is built for **10–30°**, which is the real constraint on "what lens might work here".

Found while implementing #609 / ADR 0027.

brickbots:
Closed by #652, which landed the last two gaps. Recording what happened to each part of this issue, since most of it was fixed in the interim and one bullet turned out to be stale.

**Already landed before #652:**
[...]

Issue #612, opened 2026-08-15
Calibrate the 12mm lens on-sky: effective focal length, SQM zero point, f-number


brickbots:
The 12 mm entry in `PiFinder.optics.LENSES` carries three values that are assumptions rather than measurements. One on-sky sweep settles all three.

## 1. Effective focal length (currently the nominal 12.0)

Flagged in the data so it cannot be mistaken for calibrated:

```python
"12mm": Lens(
    nominal_focal_length_mm=12.0,
    effective_focal_length_mm=12.0,
    effective_focal_length_measured=False,
    ...
)
```

The shipped 16 mm measures **15.61 mm** — 2.5% short of its label, confirmed independently on two sensors. Assuming the 12 mm is exactly 12.0 is optimistic. Everything angular derives from this value: the FOV gate, the SQM radiometric field width, and the chart's frustum shading.

Method: solve frames taken through the 12 mm and compare tetra3's **fitted FOV** against the derived one, the same comparison `tests/test_optics_solving.py` makes for the hq train. The ±15% gate is wide enough to solve through a 2.5% error, so this measures cleanly without a chicken-and-egg problem.

## 2. SQM zero-point offset (unmeasured)

`radiometric_zero_point` is calibrated per sensor against a reference meter, on the shipped lens. A different lens changes throughput, so the constant does not transfer. Currently nothing corrects for this — a 12 mm user gets the 16 mm's zero point.

## 3. f-number (currently the conservative 2.0)

Front elements measure 9.95 mm (16 mm lens) and 9.85 mm (12 mm), but the front element is **not** the entrance pupil — on M12 lenses it is deliberately oversized to accept off-axis rays. The advertised f/2 on the 16 mm implies an 8.0 mm pupil behind a 9.95 mm element, a 1.24× oversize.

Applying that same oversize to the 12 mm rules out the fast end: f/1.2 would need a pupil larger than the front element, and f/1.4 would need *less* oversize than the 16 mm, which is backwards for a wider lens. Plausible range is **f/1.6–f/2.0**; 2.0 is recorded because it never over-claims. Not load-bearing in any current code path.

Settle it photometrically rather than by measuring glass: the 16 mm runs **+0.62 mag** deeper than an f/2 12 mm but only **+0.14 mag** deeper than an f/1.6 one, against ~0.08–0.11 mag sweep repeatability — a 4–6σ separation. That is the same measurement as (2), so both close together.

## Why one sweep does all three

[...]

Issue #611, opened 2026-08-15
Gate zero-match recovery on centroid count


brickbots:
Zero-match recovery walks the exposure ladder when a solve produces no matches. Because the trigger is `Matches` and nothing checks whether the frame actually had stars in it, recovery fires on frames that are correctly exposed and full of good centroids but were rejected by the solver — and then keeps walking, because changing the exposure cannot fix a non-exposure problem.

Two ways to reach that state today:

- **Defocus** — stars smear, centroiding still finds sources, matching fails. Pre-existing.
- **A mis-stated lens** — every frame falls outside the FOV gate, so tetra3 prunes candidates before verification and `Matches` is zero on every frame regardless of exposure. New with ADR 0027, which deliberately does not self-correct a wrong lens (`docs/adr/0027-fov-gate-derived-from-optical-train.md`).

## Change

Only enter zero-match recovery when the centroid count is also low. If the extractor found plenty of sources and the solve still returned nothing, the problem is not exposure and the ladder should not run.

Prerequisite: `Centroids` on `SolveDiagnostics` (`python/PiFinder/types/positioning.py:241`), which currently carries only `Matches`, `RMSE`, `Prob`, `FOV`, `T_solve`, `T_extract`. Same prerequisite as #610 — worth landing that field once and building both on it.

The threshold wants picking against real frames rather than by inspection; the frames in `test_images/` plus the sweep archives should be enough to choose one.

Found while implementing #609 / ADR 0027.

Issue #610, opened 2026-08-15
Auto-exposure is driven by Matches, so a rejected frame looks underexposed


brickbots:
Solver-driven auto-exposure feeds `solution.diagnostics.Matches` to the PID controller (`python/PiFinder/camera_interface.py:436`). `Matches` is the count of stars tetra3 *matched* — it is zero both when the frame is genuinely too dark and when the frame is fine but the solver rejected it for some other reason.

`SolveDiagnostics` (`python/PiFinder/types/positioning.py:241`) carries no centroid count:

```python
Matches: int = 0
RMSE: Optional[float] = None
Prob: Optional[float] = None
FOV: Optional[float] = None
T_solve: Optional[float] = None
T_extract: Optional[float] = None
```

So the controller cannot distinguish "no stars in the image" from "plenty of stars, but the solve failed". A frame full of good centroids that tetra3 discards reads to auto-exposure exactly like a black frame, and the controller walks the exposure in a direction that cannot help.

This is not new — defocus has always done this — but ADR 0027 gave it a second trigger. A mis-stated lens puts every frame outside the FOV gate, so tetra3 prunes perfectly good centroids before verification and `Matches` is zero on every frame. See the "Consequences" section of `docs/adr/0027-fov-gate-derived-from-optical-train.md`.

## What to explore

Drive the controller from **centroids detected** rather than `Matches`. Centroid count measures what auto-exposure actually cares about — whether the frame has usable signal — and is independent of whether the solve succeeded.

Prerequisite: add a `Centroids` field to `SolveDiagnostics`. The count is already available in the solver at the point the diagnostics are built.

Worth deciding as part of this: whether centroid count *replaces* `Matches` as the control signal or supplements it (e.g. centroids for the exposure decision, `Matches` for a separate "are we solving?" signal).

Related: #611 gates zero-match recovery on the same new field — worth landing `Centroids` once and building both on it.

Found while implementing #609 / ADR 0027.

PR #624, opened 2026-08-17
fix(optics): FOV gate width follows lens confidence, plus lens self-heal (ADR 0029)

brickbots:
Some rev4 PiFinders shipped with a 12 mm lens and no `camera_lens` in their config. 2.6.2's derived FOV gate falls back to the imx462's `default_lens_key` of `16mm`, derives 10.40°, and gates `[8.84, 11.96]` — against frames that are actually 13.51° wide. **Those units stopped solving on update having changed nothing**, and the symptom presents as an exposure problem rather than a configuration one.

The pre-0027 window was the fixed `12.0 ± 4.0` = `[8.0, 16.0]`, which happened to span the imx462 at *both* 12 mm and 16 mm. That is the whole reason these units worked on 2.6.1 and not on 2.6.2.

> [!NOTE]
> **#625 (`lens-menu-restart`) has merged.** This branch was stacked on it; it is now rebased onto `main` with #625's commit dropped, and no longer carries it. The rebase was clean and the resulting tree is bit-identical to the pre-rebase branch, so nothing in this PR's content changed.

## The fix

A fallback is not a claim, and it should never have bought a claim's confidence. Gate width now follows how the lens was obtained:

| the lens is… | gate |
|---|---|
| **stated** — config names a lens we recognise | ±15% of that lens's field of view, exactly as ADR 0027 had it |
| **assumed** — nobody has said | the union of every lens that sensor has shipped with |

| sensor | shipped lenses | assumed gate | stated-16mm gate |
|---|---|---|---|
| imx462 / imx290 | 12 mm (13.51°), 16 mm (10.40°) | `12.19 ± 3.35` = [8.84, 15.53] | [8.84, 11.96] |
| imx296 | 12 mm (17.78°), 16 mm (13.71°) | `16.05 ± 4.39` = [11.65, 20.44] | [11.65, 15.77] |
| hq | 25 mm (10.33°) | [8.78, 11.88] — **unchanged** | n/a |

The imx462's assumed gate lands within a whisker of the pre-0027 constants it replaces, which is the behaviour these units are known to work under.

Then **self-heal closes the assumption**. A successful solve reports the fitted FOV and the sensor is already known, so what is left is the lens. Three consecutive identifications agreeing to within 5% promote the assumption to a statement, and the gate narrows onto it. It:

- **never overwrites a stated lens** — the user's claim stays authoritative, so ADR 0027's "a mis-stated lens means no solves" is untouched, and the write happens at most once per device;
- **writes nothing when the measurement matches no shipped lens** — third-party glass keeps the wide gate, which is honest;
- **skips replayed solves** — a recorded session measures whatever device recorded it;
- **can never take the integrator down** — a read-only filesystem or a dropped manager connection costs the promotion, not the pointing.

## Why a union rather than dropping the hint

Measured against the real solver and real frames before committing to this: **widening costs nothing on speed.** ±15%, ±30%, ±50% and the pre-0027 fixed window all solve in 1.1–1.4 ms, and even no hint at all costs only +0.3 ms; `Prob` was bit-identical across every width.

But the gate **does** buy mis-solve rejection. With injected noise, wide and hintless gates returned *confident* false solves at 20–23° that `match_threshold` did not reject. The assumed gate still excludes everything above 15.53°, so it keeps that protection — omitting `fov_estimate` would give it up. (Single-shot noise trials: direction established, magnitude not.)

And the cost is wildly asymmetric. A good frame *outside* the gate burns the entire `solve_timeout` and returns nothing, every frame, forever. Against a wide gate's ~0.3 ms, an over-tight gate is the expensive failure by three orders of magnitude.

## Rejected

- **A rev4 → 12 mm config migration.** Rev4 reports as `imx462` and so do some v3 units, so the sensor is not the predicate. The board is probeable, but board revision is not the fact we need — *which lens was in the box* is, and not all rev4s shipped with the 12 mm. It would fix most units by breaking the rest, and break them in exactly the way this PR exists to fix.
- **`camera_lens` in `default_config.json`.** `Config.get_option` consults the default file before `resolve_lens` ever sees a `None`, so hq would derive 17.12° instead of its true 10.33°.
- **Raising `FOV_GATE_MARGIN` globally.** Spends 0027's entire benefit on every correctly-configured user to cover a case they are not in.

Full reasoning and the numbers: `docs/adr/0029-fov-gate-width-follows-lens-confidence.md`.

## The hq must not move

The hq only ever shipped one lens, so it has no union to take. That case returns early rather than falling through the union arithmetic — which re-derives the same window through different float operations and lands 1 ULP off the value that has been shipping. `test_the_hq_gate_is_bit_identical_to_what_shipped` asserts the exact floats, not `approx`.

## Verification

- **1,285 unit + smoke tests pass**, 0 failures. Ruff lint + format clean, mypy clean over 151 files.
- **115 new test functions** across `test_optics.py` (63) and the new `test_lens_self_heal.py` (19), plus the existing suites.
- **Sphinx builds with zero warnings** (`-n -q`), and the new prose carries no em-dashes or semicolons.
- **Self-heal verified end to end on the running app** (`--camera debug`), which exercises the write path through the integrator wiring: `Lens self-heal: 3 solves fitted 10.20 deg, which is the 25mm (10.33 deg derived).` The dev config was backed up and restored to its original checksum afterwards.

> [!WARNING]
> The debug camera declares `hq`, which has one shipped lens, so its assumed gate is identical to its stated one. **The debug run cannot exercise the widening at all** — the imx462 12mm-vs-16mm case, which is the entire point of the change, is only reachable in unit tests or on real hardware. A green debug run is not evidence the fix works on an affected unit.

## Known limitation, not fixed here

After self-heal writes `camera_lens`, the UI process's `Config` still holds the dict it loaded at boot. Any later `set_option` from the menu rewrites `config.json` from that stale dict and drops the self-healed key. This is **self-correcting** — the device re-heals after three solves on the next boot, and `shared_state` carries the correct lens for the whole session, so solving is never affected. It is also pre-existing: `camera_interface.py` already writes config from a non-UI process with the same exposure. Worth a follow-up issue rather than a wider change to `Config` in this PR.

## Docs

- `release_notes/2.6.2.md` — the fix, and two corrections it forces: the release no longer claims "no configuration file is rewritten", and the "device deliberately does not recover" sharp edge is restated (self-heal writes into an absence, never over a statement). A new sharp edge covers the post-self-heal physical lens swap.
- `docs/source/troubleshooting.rst` (#613) — the "changed lens → no solves" path, narrower now that self-heal covers the never-stated case.
- `docs/source/BOM.rst` (#613) — "something fast with a 10deg FOV" was imx462-specific and contradicted the parts list two lines above it.
- ADR 0029's imx296 row: `[11.66, …]` → `[11.65, …]`, a rounding slip. 11.65 is what the release notes already quoted.

Closes #613. Related: #610, #611, #612.


brickbots:
Reviewed this branch. Verified the PR's own claims independently: 1,285 unit + smoke tests pass, `ruff check`/`format` clean, mypy clean on the changed modules, and `sphinx -n -q` builds with zero warnings. The stated/assumed split and the union arithmetic both check out against the ADR table, and the hq early-return with the bit-identical float assertion is the right call.

One defect found and pushed as 5514a39 (now 1,286 passing):

**The Lens menu showed the wrong lens after self-heal, and confirming it was unrecoverable.**

`get_camera_lens` resolved the sensor half from `shared_state` but the lens half from `config_object`. The UI process loads its `Config` at boot and reloads only on an explicit `reload_config`, so after self-heal writes from the integrator the menu still reported the *assumed* lens. Confirmed against this branch:

```
after self-heal, on disk : 12mm
Lens menu displays       : 16mm   (shared_state says 12mm)
```

Not only cosmetic: `text_menu` writes the highlighted entry on select, so a user confirming that stale `16mm` on a healed 12mm unit states the wrong lens, tightens the gate to `[8.84, 11.96]` against 13.51° frames, and stops solving permanently — and self-heal can't recover it, since it only writes into an absence. The `troubleshooting.rst` added by this PR sends a user with no solves to exactly that screen.

The fix reads `shared_state.camera_lens()` instead. Shared state is seeded from config at construction and republished by both `set_camera_lens` and self-heal, and nothing else writes `camera_lens` (`server.py` never touches it), so it is strictly the more current view. Added `test_shows_the_self_healed_lens_not_this_process_stale_config`, which fails on the old code with `['16mm'] != ['12mm']`.

Filed #626 for the general form of the staleness — `dump_config` rewrites the whole file from a per-process dict, so a cross-process write is dropped by the next UI write. Worth reading before merge for one reason: it means **"it writes at most once per device" is not accurate** as currently written in ADR 0029 and `release_notes/2.6.2.md`. A device with an unstated lens re-heals every boot. Harmless (shared state carries the right lens all session) but the claim should be softened. `server.py` already has the mechanism to fix it properly — `ui_queue.put("reload_config")` — the integrator just isn't passed `ui_queue`.

Smaller points, not pushed:

- **Self-heal will write a lens identified against a *fallback* camera profile.** `resolve_camera_profile` silently falls back to imx296 for an unknown sensor, and `test_an_unrecognised_sensor_does_not_raise` asserts that as desired. If the sensor isn't really an imx296 the identified lens is wrong, and the write is unrecoverable — which contradicts the ADR's own "a wrong write is worse than no write". Near-unreachable in practice (`camera_interface` sets `camera_type` before the first capture), but the guard is one line: skip when `camera_type not in CAMERA_PROFILES`.
- **The imx296 assumed gate reaches 20.44°**, inside the 20–23° band where the ADR reports wide gates returned confident false solves. The "still excludes everything above 15.53°" argument — and `test_the_assumed_gate_still_rejects_a_wild_field_of_view` — are imx462-specific. Worth a sentence in the ADR, or an imx296 bound in the test.
- **`except (OSError, AttributeError, ValueError)`** is narrower than the docstring's promise: a dropped manager connection usually surfaces as `EOFError`, which isn't an `OSError`. And catching `AttributeError` will silently disable self-heal on a rename rather than failing loudly. A plain `except Exception` with the same disable-and-warn matches the stated intent and swallows nothing more.
- **No test pins `LENS_IDENTIFY_TOLERANCE < FOV_GATE_MARGIN`**, which is what guarantees a mis-identification still solves inside the gate the write produces. Load-bearing, one line.
- **`LensSelfHeal.__init__` is untyped**, so mypy skips its body (`integrator.py:116: note: bodies of untyped functions are not checked`). `_write(..., profile)` likewise.
- **`docs/source/menu_map.rst` adds an em-dash** ("what is fitted — and naming the wrong one"), which the published-docs house style replaces; the PR body says the new prose carries none. ADR/CONTEXT em-dashes are fine. `BOM.rst`'s addition also ends without a full stop and writes `12mm` where the release notes use `12 mm`.

Nothing else blocking from me.

PR #633, opened 2026-08-25
feat(chart): name the center object on an optional readout

brickbots:
Implements the chart **center-object readout** from the settled design: an optional line along the bottom of the chart naming the **center object** — the plotted marker nearest the center of the chart — with a right-arrow glyph advertising that RIGHT opens that object's details.

Reasoning lives in the two docs that ship with this PR: **ADR 0031** (why the candidate set is the drawn markers rather than the Nearby ranking, why distance is in screen pixels, why the choice is sticky) and the new **Chart** section of `docs/ax/ui/CONTEXT.md` (which also flags *center object* vs *target*).

## What's here

- **Candidate set** — `plot_markers` now records what it drew (`_collect_dso_markers` returns the objects behind the marker list). The target cross first when it has RA/Dec, then the DSO layers. With DSO Display off the existing early return leaves the cross as the only candidate, for free.
- **`CenterObjectTracker`** (`ui/center_object.py`) — drops off-screen markers, ranks by pixel distance from the chart center, and holds the incumbent unless a rival is closer by `STICKY_MARGIN` (0.15). Identity is `object_id`; it exposes whether the pick changed, which drives the marquee reset.
- **The readout** — `TextLayouterScroll` at the user's `text_scroll_speed`, rebuilt when the center object changes. Stacks from the bottom: RA/Dec keeps the bottom line when on, otherwise the readout takes it. Recomputed each frame, since the setting can change while the preloaded chart is live.
- **RIGHT** — opens the center object's details, with the nearest-first on-screen candidates as `object_list` so UP/DOWN walks outward through the chart's own markers. Gated on the readout; off/blank/no-solve leaves it inert.
- **`chart_center_object`** (Off/On, **default On**) as a `Center Object` item in Settings > Chart. Note the non-obvious bit: existing charts gain a line on upgrade. That's deliberate.
- **`UIChart.serialize_ui_state`** — `fov`, `center_object_readout`, and a `center_object` dict, so the feature is checkable without OCR-ing a screenshot.
- User guide + menu map, and the `Center Object` string in all four `.po` files.

## Why the tracker is its own module

`ui/chart.py` carries `# mypy: ignore-errors`, and chart-level tests need `hip_main.dat`, which is git-ignored and doesn't ship. Keeping the tracker free of PIL and `Starfield` makes it both unit-testable and type-checked. 21 new unit tests cover nearest-wins, hysteresis in both directions, the margin scaling with zoom, off-screen exclusion, half-open bounds, the empty set, the incumbent vanishing, and the string builder's dedup.

## The frame-rate constraint

The chart repaints the starfield only on a new solve, but `screen_update()` pushes a frame every loop iteration. A marquee driven off the new-solve path would stutter at 1–2 Hz. So the finished chart is cached after `draw_reticle()`, and every frame repaints the strip from that backdrop before drawing the text. The backdrop is dropped in `plot_no_solve`, which clears the screen.

## Verified on the running app

Launched headless with `--camera debug` and driven through `pf_remote`:

- Readout renders above the RA/Dec line with the arrow pinned right; and takes the bottom line with RA/DEC Disp. off.
- RIGHT opened M 57's details with `object_list_length: 4`; LEFT returned to a live readout.
- `Center Object` appears in Settings > Chart with On checked by default; setting Off blanks the line and makes RIGHT inert (`center_object_readout: false`).
- `/api/current-selection` reported e.g. `{"center_object": {"display_name": "M 57", "object_id": 6720, "screen_px": [40.3, 58.4]}, "center_object_readout": true, "fov": 10.2}`.

`nox -s lint format type_hints` clean (mypy: 152 files), 1329 unit + 5 smoke tests pass, docs build with `-n` and no warnings.

## Worth a look before merge

**The "first common name" is usually a cross-designation, not the popular name.** The design said designator + first entry of `name_deduplicate(obj.names, [obj.display_name])`, and that is what this does — but against the real catalog it reads:

| Object | Readout |
|---|---|
| M 45 | `M 45 Cr 42` — the catalog has no "Pleiades" entry at all |
| M 57 | `M 57 NGC 6720` |
| M 31 | `M 31 NGC 224` |
| M 13 | `M 13 NGC 6205` |
| NGC 7000 | `NGC 7000 North America nebula` |

`names` is in catalog order and `UIObjectList` lists the same names in the same order, so this is consistent with the rest of the UI rather than a bug here — but the worked example in the handoff ("M 45 Pleiades") isn't reachable. Preferring a non-designation name would be a change to that shared convention, so I left it as specified, documented the consequence at `center_object_text`, and used a truthful example in the user guide. Say the word if you'd rather it skipped cross-designations.

**Tuning.** `STICKY_MARGIN = 0.15` and the exact strip geometry both want a real device before they're final; the constant sits next to `NEARBY_MARKER_CAP` with the same "tune on device" note.

**Translations** are AI-drafted and marked `AI-TRANSLATED (claude)` for review: de `Zentralobjekt`, es `Objeto Central`, fr `Objet Central`, zh `中心天体`. The `.po` diff is 7 lines per language — `messages.pot` is git-ignored here, so I hand-added the entry rather than running a full extract that would churn every line reference.

**Follow-up, not in this PR:** the on-device help screens under `help/chart/` are authored PNGs and still advertise no RIGHT action for the chart.

https://claude.ai/code/session_01EUnENdJJncdHkfvdMTxEk4


brickbots:
Pushed `a3ec3f5a` addressing findings 1–3 from the review.

**1. Batched the projection.** `Starfield.radec_to_xy` builds a one-row DataFrame, `Star` and `observe()` per call, and `_update_center_object` looped it over every drawn marker — about 21 skyfield round trips per solve, re-deriving coordinates `plot_markers` had just computed vectorized. New `radec_to_xy_many` does the whole candidate set in one round trip; `radec_to_xy` delegates to it, so the scalar callers in `align.py` and `_draw_target` are unchanged. Measured over 21 markers: **3.68 ms → 0.18 ms per solve (20.8×)**, output bit-identical to the scalar path (max difference exactly 0.0 across 26 positions, including RA wrap-around and both poles, with a non-zero chart roll).

**2. Skipped the work when the readout is off.** `_update_center_object` was unconditional, so users with Center Object set to Off paid the full projection every solve for a pick nothing could show — RIGHT is gated on the readout too. It now returns early and clears the tracker.

This does change the published state: `center_object` is now `null` whenever `center_object_readout` is `false`, instead of naming an object the user can neither see nor open. Flagging it since the original docstring promised the opposite.

**3. Moved the pure helpers into the testable module.** `readout_y`, `readout_scroll_speed` and `readout_enabled` had no need of PIL or `Starfield` — only config values and two integers — so they were on the wrong side of the boundary `ui/center_object.py` exists to draw. They're free functions there now, with `ui/chart.py` keeping the config reads that feed them. New tests cover the bottom-stacking rule against both RA/Dec formats, the strip never crossing the RA/Dec line, the fallback when `text_scroll_speed` is unrecognised (Medium, not 0 — falling back to "no scroll" would silently clip long names), and layout on a 320px panel.

**Verified**

- 1346 unit tests pass (up from 1329), 5 smoke tests pass; `ruff check`, `ruff format --check` and `nox -s type_hints` (152 files) all clean.
- Driven on the running app headless with `--camera debug`: readout renders above the RA/Dec line with the arrow pinned right (`M 57 NGC 6720`); `/api/current-selection` reports e.g. `{"center_object": {"display_name": "M 57", "object_id": 6720, "screen_px": [27.4, 46.0]}, "center_object_readout": true, "fov": 10.2}`; RIGHT opens the center object's details; with Center Object set to Off the line disappears with no stale text left on the strip and the API reports `"center_object": null, "center_object_readout": false`; toggling back On restores it.

Findings 4–6 from the review (the `ranked[0]` wording when hysteresis holds, the full-frame backdrop copy, `int()` vs `ceil` on the arrow width) were minor and are left alone. The `help/chart/` PNGs still advertise no RIGHT action — worth landing in the same release even if not this PR.

PR #634, opened 2026-08-25
docs(network): Network bounded context, ADR 0031, and the secure-WiFi refresh plan (EN 18031)

brickbots:
## What this is

WS1 of the plan to refresh **#311** ("[BETA] More secure WiFi configuration", implements **#179**) and get it released. Docs only — the code refresh follows as its own PR, reviewed against the vocabulary and decisions this PR puts on record.

## Why now

EN 18031 (the EU RED cybersecurity regulations) requires new units to ship with an **encrypted AP and a per-unit random WiFi password**. That reverses the opt-in posture #311 settled on in 2025 — and, notably, lands on the posture #179 originally asked for. The decision, its history, and its consequences are recorded in **ADR 0031** so the compliance trail exists before any code changes.

## Contents

- `docs/ax/network/CONTEXT.md` — new Network bounded context: WiFi mode, Access Point, provisioning tokens (`CHANGEME`/`ENCRYPTME`), **new unit vs field-upgraded unit** (the compliance boundary — upgrades never force-encrypt), AP passphrase, Connect WiFi screen, WiFi QR code, recovery reset.
- `docs/adr/0031-ap-provisioning-ssid-tokens-encrypted-default.md` — why tokens live in the SSID, why the default flipped, considered options, consequences (including: the recovery config must **not** carry `ENCRYPTME`, or a recovered unit re-locks itself on the next boot).
- `CONTEXT-MAP.md` — Network context entry + relationships (Network → UI, Network → NixOS migration seam).
- `secure-wifi-refresh-plan.md` — the full release-readiness plan: drift analysis of #311 vs today's main (Bottle→Flask port, docs restructure, three panel geometries, i18n), the defect list found reading the diff, and workstreams WS2–WS8 through hardware validation.

## Open questions (deliberately not settled — grill fodder)

1. May the web UI disable encryption on a provisioned unit? (leaning no)
2. Do the two tokens collapse into one? (leaning no — `ENCRYPTME` alone is the opt-in path for field units)
3. Should first boot surface the Connect WiFi screen unprompted? (#179 asked for this)
4. Does the physical-access recovery reset satisfy EN 18031? To be confirmed and recorded in ADR 0031.

## Side finding

`CONTEXT-MAP.md` already referenced `docs/ax/nixos/CONTEXT.md` and `docs/ax/nixos.md`, neither of which exists on main. Not touched here — worth a separate tiny fix.

Refs #179, #311.

https://claude.ai/code/session_01W6UPq3yDJTqti9kgQXA7NZ


mrosseel:
Needed work, had been thinking about it when doing the recent file explorer in the web and also some web refactoring. Wifi qrcode would be great for discover ability. 
Code should be done on the Nixos branch though (or a pr on it), will be a big change and network specifically touches Nixos config. Willing to look at it if specs are clear

brickbots:
Thanks @mrosseel!  I was indeed hoping this could wait until after the NixOS migration, but it seems like it's a blocking requirement for selling PiFinders in the EU.  Maybe we should have an offline discussion about the timeline for NixOS migration as there are some things that may help compliance simply by doing the migration 🤔

